PR #88, opened 2023-10-10
feat: Update test vectors

seebees:
The MPL test vectors support CMMs and have extensive tests for the DefaultCMM and the CachingCMM.
This also removes some tests that were not running
as well as refactoring the test vectors to be more manageable for a single object under test. 


fpawlega:
+ add an explicit test case that asserts required keys are not serialized into the header by parsing the header wire format
+ correct any test cases from the tabular test definitions that had incorrect expected behaviour

PR #99, opened 2023-10-19
chore: Consolidate tasks

seebees:
* Update the root Makefile to run for each library
* Add format and codegen targets to root Makefile
* Add java and net format/check targets
* Update format action to prepare for update

### DIFF
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   manual-ci-format:
-    uses: ./.github/workflows/library_dafny_format.yml
+    uses: ./.github/workflows/library_format.yml
     with:
       dafny: '4.2.0'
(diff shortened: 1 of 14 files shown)


texastony (dismissed):

.github/workflows/manual.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   manual-ci-format:
-    uses: ./.github/workflows/library_dafny_format.yml
+    uses: ./.github/workflows/library_format.yml
     with:
       dafny: '4.2.0'
```
texastony, line 18: We need to pull these Dafny Version variables out to a common definition!

texastony (approved):

.github/workflows/manual.yml
seebees: Agree.

PR #116, opened 2023-10-31
chore: Update gradle projects

seebees:
Use Composit builds over `mavenLocal()`

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
+++ b/TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
@@ -50,7 +50,6 @@ if (!caPasswordString.isNullOrBlank()) {
 
 repositories {
     mavenCentral()
-    mavenLocal()
     if (caUrl != null && caPassword != null) {
         maven {
             name = "CodeArtifact"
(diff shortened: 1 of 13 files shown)


TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
```diff
@@ -67,6 +66,10 @@ dependencies {
     implementation("org.dafny:DafnyRuntime:${dafnyVersion}")
     implementation("software.amazon.smithy.dafny:conversion:0.1")
     implementation("software.amazon.cryptography:aws-cryptographic-material-providers:1.0.1")
```
texastony, line 68: Question: Why not SNAPSHOT for the MPL?

texastony (approved):

TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
seebees: Because the goal of the TestVectors is to be published and match the MPL version.
This means that for local development it gets its settings from `settings.gradle.kts`
But it can eventually be run and is explicitly bound to a specific version of the MPL

PR #127, opened 2023-11-16
fix: Empty string defers to SDK default region

seebees:
The ESDK will pass an empty string
as the AWS region in cases where no region is know.

In this case we defer to the SDK
on how to evaluate the default region.


josecorella (approved):
lgtm

PR #131, opened 2023-11-28
feat: add command line parser

ajewellamz:
_Issue #, if available:_

_Description of changes:_


StandardLibrary/src/GetOpt.dfy
```diff
@@ -0,0 +1,435 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+include "../../../../submodules/MaterialProviders/libraries/src/Wrappers.dfy"
+include "../../../../submodules/MaterialProviders/StandardLibrary/src/StandardLibrary.dfy"
```
seebees, line 5: I don't think these are the right path?

StandardLibrary/src/GetOpt.dfy
```diff
@@ -0,0 +1,435 @@
...
+        else
+          GetMaps(opts[1..], longMap[opt.name := opt.Opt?], shortMap, commandMap)
+  }
+
+  function printFromFunction<T>(x: T): () {
+    ()
+  } by method {
+    print x,"\n";
+    return ();
+  }
```
seebees, line 271: I do this too, I was thinking 
```suggestion
  function Print<T>(x: T): Outcome<string> {
    Pass
  } by method {
    print x,"\n";
    return Pass;
  }
```

Then you can
```dafny
:- Print("less typing!");
```

StandardLibrary/src/GetOpt.dfy
```diff
@@ -0,0 +1,435 @@
...
+
+  datatype Context = Context (
+    longMap : map<string, bool>,
+    shortMap : map<char, string>,
+    commands : CommandMap,
+    command : string
+  )
+
+  function /*{:tailrecursion}*/ GetOptions2(
```
seebees, line 290: What is the diff in the 2?

ajewellamz: GetOptions is the thing the user calls.  GetOptions2 is the recursive thing that implements most of GetOptions.

seebees (approved):
Love it.
Much more complete that what I hand rolled.
Thanks.

PR #132, opened 2023-11-30
feat(STD-NET): support netstandard2.0

texastony:
_Issue #, if available:_ https://github.com/aws/aws-encryption-sdk-dafny/issues/620

_Description of changes:_
- Add `netstandard2.0` to all publishing csprojs 
- Enable Dependabot for .NET

`chore(NET): support netstandard2.0`

## Testing:
See conversation b/w @josecorella and @texastony .
We struggled to get net core app 2.0 testing in CI.
Locally, we ran some net core app 2.0 testing.

We MAY want to get net core app 3.1 or another target into CI.
Or, we can trust Nuget/.NET's tooling to assert that we only use APIs
supported by `netstandard2.0`. 

(Since we only use supported APIs,
and APIs SHOULD behave the same across Frameworks,
our code SHOULD behave the same across Frameworks,
and additional testing is not needed.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,33 @@ updates:
     directory: ".github/workflows"
     schedule:
       interval: "daily"
+  # STD NET
+  - package-ecosystem: "nuget"
+    directory: "StandardLibrary/runtimes/net/STD.csproj"
+    schedule:
+      interval: "daily"
+  # KMS NET
+  - package-ecosystem: "nuget"
+    directory: "ComAmazonawsKms/runtimes/net/AWS-KMS.csproj"
+    schedule:
+      interval: "daily"
+  # DDB NET
+  - package-ecosystem: "nuget"
+    directory: "ComAmazonawsDynamodb/runtimes/net/ComAmazonawsDynamodb.csproj"
+    schedule:
+      interval: "daily"
+  # Crypto NET
+  - package-ecosystem: "nuget"
+    directory: "AwsCryptographyPrimitives/runtimes/net/Crypto.csproj"
+    schedule:
+      interval: "daily"
+  # MPL NET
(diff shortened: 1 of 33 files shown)


texastony:
@josecorella and I had a call about testing `netstandard2.0`.

While we disagreed on much, 
we ended up agreeing that because Net Framework 4.8.1
is supported by net standard 2.0 but not 2.1,
if any of our dependencies moved from net standard 2.0 to net standard 2.1,
our CI would fail.

Thus, we do not need CI testing that explicitly tests for net standard 2.0 only targets.

There is a case I had not thought of:
What if a dependency drops net standard entirely?

i.e: Bouncy Castle goes from supporting net standard 2.0, to just net481.
Our CI would all pass, since we test against net6.0 and net481.

But would dotnet give some error or warning that our dependency 
conflicted with our target frameworks?

(And would we notice such an error or warning, 
if it was just emitted in CI?)

texastony:
More From @josecorella :
>if it removes netstandard2.0 and there is no supported framework equal to what we ship, yes it will fail
because the build command will error with something like “configured version does not match downstream dependency “x.x.x”
> or a similar error

At this point, I know the tests pass on my local with net core app 2.0,
and that it is proving difficult to test net standard 2.0 only targets.

I am going to remove the net core app 2.0 test actions,
and then this PR should be good to go.

texastony:
It turns out I clearly have not run all the projects against net core app 2.0 locally,
as some of the test projects were never modified...

josecorella (changes requested):
we need to update the signing and release process to account for the new target we are adding. 
@texastony, we can chat to see what work if any needs to be done

ajewellamz (approved):
LGTM

josecorella (approved):

PR #206, opened 2024-02-13
chore(GHA): add semantic release manual gha

josecorella:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/.github/workflows/sem_ver.yml
+++ b/.github/workflows/sem_ver.yml
@@ -47,5 +47,5 @@ jobs:
       # Log in as the CI Bot
       - name: Log in as CI Bot
         run: |
-            gh auth login --with-token < echo "${{env.GITHUB_AWS_CRYPTO_TOOLS_CI_BOT_PERSONAL_ACCESS_TOKEN__NEW_TOKEN_FORMAT_}}"
+            gh auth login --with-token < "${{env.GITHUB_AWS_CRYPTO_TOOLS_CI_BOT_PERSONAL_ACCESS_TOKEN__NEW_TOKEN_FORMAT_}}"
             gh auth status
--- a/.github/workflows/sem_ver.yml
+++ b/.github/workflows/sem_ver.yml
@@ -47,6 +47,6 @@ jobs:
       # Log in as the CI Bot
       - name: Log in as CI Bot
         run: |
-            echo ${{env.GITHUB_AWS_CRYPTO_TOOLS_CI_BOT_PERSONAL_ACCESS_TOKEN__NEW_TOKEN_FORMAT_}} > token.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/sem_ver.yml
```diff
@@ -0,0 +1,55 @@
...
+      # Log in as the CI Bot
+      - name: Log in as CI Bot
+        run: |
+          echo ${{env.GITHUB_AWS_CRYPTO_TOOLS_CI_BOT_ESDK_RELEASE_TOKEN}} > token.txt
+          gh auth login --with-token < token.txt
+          rm token.txt
+          gh auth status
+
+      # Test to see if we can setup semantic release
+      - name: Run Semantic Release in dry run mode
```
lucasmcdonald3, line 50: Are we just setting it up? Or also running dry run?

josecorella: oops forgot to change the name of the step; fixed now!

.github/workflows/semantic_release.yml
```diff
@@ -0,0 +1,70 @@
+# This workflow runs semantic release, bumps, generates changelog, and tags the project
+name: Semantic Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Are you releasing a new version? (Y/n)"
```
lucasmcdonald3, line 8: I think the convention for CLIs is that a capital letter suggests it's a default argument. Should `Y` be lowercase `y` here?

josecorella: yup!

.github/workflows/semantic_release.yml
```diff
@@ -0,0 +1,70 @@
...
+        required: true
+        type: string
+
+jobs:
+  semantic-release:
+    if: contains('["seebees","texastony","ShubhamChaturvedi7","lucasmcdonald3","josecorella","imabhichow","rishav-karanjit","antonf-amzn","justplaz","ajewellamz"]', github.actor)
```
lucasmcdonald3, line 14: Is it worth considering something like https://github.com/tspascoal/get-user-teams-membership?

.github/workflows/sem_ver.yml
```diff
@@ -0,0 +1,55 @@
+# This workflow runs semantic release in dry run mode
```
lucasmcdonald3, line 1: Is this comment still accurate?

.github/workflows/semantic_release.yml
josecorella: that would be nice but unfortunately it doesn't look like it is a verified gha; otherwise I would use it

lucasmcdonald3: Make sense.
Can we add a comment that points to the team this is expected to be synced with here?

lucasmcdonald3 (approved):

PR #225, opened 2024-02-27
MPL-Core: Smithy-generated code

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_


AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
+# Code generated by smithy-python-codegen DO NOT EDIT.
+
+import asyncio
+from asyncio import sleep
+from copy import deepcopy
+from software_amazon_cryptography_keystore_internaldafny_types import IKeyStoreClient
```
ShubhamChaturvedi7, line 6: Do you have a separate branch with `.` replaced with `_` in the dafny source code?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+class KeyStore:
+    """Client for KeyStore
+
+    :param config: Configuration for the client.
+    """
+
+    def __init__(
+        self,
+        config: KeyStoreConfig | None = None,
+        dafny_client: IKeyStoreClient | None = None,
```
ShubhamChaturvedi7, line 67: Do we need to expose this internal dafny class to the customers?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+        deserialize: Callable[[DafnyResponse, Config], Awaitable[Output]],
+        config: Config,
+        operation_name: str,
+    ) -> Output:
+        try:
+            return await self._handle_execution(
+                input, plugins, serialize, deserialize, config, operation_name
+            )
+        except Exception as e:
+            # Make sure every exception that we throw is an instance of ServiceError so
```
ShubhamChaturvedi7, line 214: Why wrap everything in `ServiceError`? Is it the language idioms?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+        )
+        interceptors = client_interceptors
+
+        try:
+            # Step 1a: Invoke read_before_execution on client-level interceptors
+            for interceptor in client_interceptors:
+                interceptor.read_before_execution(context)
+
+            # Step 1b: Run operation-level plugins
+            # config = deepcopy(config)
```
ShubhamChaturvedi7, line 248: nit: Commented code.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+            for plugin in plugins:
+                plugin(config)
+
+            _client_interceptors = config.interceptors
+            interceptors = cast(
+                list[Interceptor[Input, Output, DafnyRequest, DafnyResponse]],
+                _client_interceptors,
+            )
```
ShubhamChaturvedi7, line 256: Why do we need this again?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+                if isinstance(context_with_response.response, Exception):
+                    # Step 7u: Reacquire retry token if the attempt failed
+                    try:
+                        retry_token = retry_strategy.refresh_retry_token_for_retry(
+                            token_to_renew=retry_token,
+                            error_info=RetryErrorInfo(
+                                # TODO: Determine the error type.
+                                error_type=RetryErrorType.CLIENT_ERROR,
+                            ),
+                        )
```
ShubhamChaturvedi7, line 323: We shouldn't retry here. It's a local call and not a service call so it's highly unlikely that an error is a transient one.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+    ) -> InterceptorContext[Input, Output, DafnyRequest, DafnyResponse | None]:
+        try:
+            # Step 7a: Invoke read_before_attempt
+            for interceptor in interceptors:
+                interceptor.read_before_attempt(context)
+
+            # Step 7m: Involve client Dafny impl
+            if config.dafnyImplInterface.impl is None:
+                raise Exception("No impl found on the operation config.")
+
```
ShubhamChaturvedi7, line 361: Should we move this up the chain?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
...
+            except Exception as e:
+                context._response = e
+
+        # Step 12: Return / throw
+        if isinstance(context.response, Exception):
+            raise context.response
+
+        # We may want to add some aspects of this context to the output types so we can
+        # return it to the end-users.
+        return context.response
```
ShubhamChaturvedi7, line 462: It'd be great if you can link any doc explaining the steps. I assume this comes from smithy-python and not the smithy-dafny codegen.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/shim.py
```diff
@@ -0,0 +1,211 @@
...
+            message=e.message, list=e.list
+        )
+
+    if isinstance(e, OpaqueError):
+        return software_amazon_cryptography_keystore_internaldafny_types.Error_Opaque(
+            obj=e.obj
+        )
+
+
+class KeyStoreShim(
```
ShubhamChaturvedi7, line 80: I thought we were only generating shims for Tests. Do we need this in implementation?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/references.py
```diff
@@ -0,0 +1 @@
+# Code generated by smithy-python-codegen DO NOT EDIT.
```
ShubhamChaturvedi7, line 1: Remove if not required.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/references.py
```diff
@@ -0,0 +1,652 @@
...
+    OnDecryptOutput_OnDecryptOutput as DafnyOnDecryptOutput,
+    OnEncryptInput_OnEncryptInput as DafnyOnEncryptInput,
+    OnEncryptOutput_OnEncryptOutput as DafnyOnEncryptOutput,
+    PutCacheEntryInput_PutCacheEntryInput as DafnyPutCacheEntryInput,
+    UpdateUsageMetadataInput_UpdateUsageMetadataInput as DafnyUpdateUsageMetadataInput,
+)
+from typing import Any, Dict
+
+
+class IBranchKeyIdSupplier(metaclass=abc.ABCMeta):
```
ShubhamChaturvedi7, line 37: What is ABCMeta?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/references.py
```diff
@@ -0,0 +1,652 @@
...
+
+    @classmethod
+    def __subclasshook__(cls, subclass):
+        return hasattr(subclass, "GetBranchKeyId") and callable(subclass.GetBranchKeyId)
+
+    @abc.abstractmethod
+    def get_branch_key_id(
+        self,
+        param: "aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.GetBranchKeyIdInput",
+    ) -> "aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_materialproviders.models.GetBranchKeyIdOutput":
```
ShubhamChaturvedi7, line 47: Why the return type is in `"`?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: No, it's not a separate branch.
I'm keeping any `.`/`_` things out of scope until I get to per-module outer-module names.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: No, but Python classes are only allowed one `__init__` method. There may be some way to refactor this to create a static method on generated clients like `from_dafny_client".

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: This is how Smithy-Python generates its clients

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: I don't know if we need this. This is generated by Smithy-Python. I've left it in internal-only code, but sealed off all the "entry points" to these interfaces. This might be useful at some point, but I tend to doubt it. If we remove this later on, it would not be a breaking change.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: Correct, we aren't retrying. This structure comes from Smithy-Python. I've generated a no-retry, no-delay token for use here: https://github.com/aws/aws-cryptographic-material-providers-library/pull/225/files#diff-cdf1f2f63adc592b23747994bd05ead4ec705088b8b4d0f36df2808485d073e1

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: I don't think there's a doc unfortunately -- it comes from Smithy-Python.
It is complicated, but at a high level, the important steps are
1. Serialize the native input into a Dafny input
2. Call the Dafny client with the Dafny input
3. Deserialize the Dafny output to a native output
4. If the native output is an exception, raise the output; otherwise, return the output.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/references.py
lucasmcdonald3: This is a "forward reference": https://peps.python.org/pep-0563/
It allows typehints for types that have not yet been defined (ex. if the type is defined later in a file, a typehint referencing that type will fail as the type has not been defined. A forward reference tells Python that it will exist at some point.)
For this particular input type, the forward reference isn't needed. But the codegen doesn't know which input types need forward references and which don't; it liberally applies forward references to avoid issues with typehints.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/references.py
lucasmcdonald3: This module: https://docs.python.org/3/library/abc.html
"abc" -> "abstract base class". It's a Pythonic way to write interfaces.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: We could, but there's no strong need to do this. This is a weird edge case that customers should never see; as long as this happens before it's called, it should be fine

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/shim.py
lucasmcdonald3: You're right -- these should only be in tests.
I included these to demonstrate the codegen for them. I had a note in my tracking doc to remove them before launch.
However, I just realized the implementation depends on the `_smithy_error_to_dafny_error` in this file.
It should be easy to move this. I've made a note to refactor this in my tracking doc.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: Removed

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/references.py
lucasmcdonald3: Done, had to move some stuff around in Polymorph

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/references.py
ShubhamChaturvedi7: So it's like a shorthand for an abstract class. Nice.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: Yeah I got at it a very high level but wasn't sure of the implementation level details, more so because this is directed towards a remote service. Maybe we should add a comment that this is from smithy-python?

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: Is it being checked somewhere? I might have missed it.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: I think I worded the comment wrong. I meant that this same code block is repeated twice.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: I think with the whole outer-module stuff, customers might not have access (shouldn't access) to the internal dafny stuff.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: So it's a local change and you are using that to generate this code?

lucasmcdonald3: Oh, I guess there actually is a separate branch, which is just the "authoritative" branch for these changes: https://github.com/aws/aws-cryptographic-material-providers-library/tree/lucmcdon/python-mpl

Ex. I've committed changes like this to the Dafny source code: https://github.com/aws/aws-cryptographic-material-providers-library/blob/lucmcdon/python-mpl/AwsCryptographyPrimitives/src/Index.dfy

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: Correct, they shouldn't do this, even regardless of outer-module changes.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: Oh! Then the answer is:
I suspect this is useful for Smithy-Python in case a plugin modifies `config.interceptors` in the `plugin(config)` call a few lines above.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: A long comment that explains how this integrates with Smithy-Python makes sense.
I don't think that should go in this file, though. I'll put that in the Smithy-Dafny code.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
ShubhamChaturvedi7: Ahh that's where `None` type comes in play. Got it.

ShubhamChaturvedi7 (approved):
Overall changes LGTM. But a second pair of eyes are always useful in long PRs :)

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
lucasmcdonald3: I guess this wasn't actually removed...
I'll address this in a follow-up PR

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/client.py
```diff
@@ -0,0 +1,884 @@
...
+    _serialize_validate_commitment_policy_on_encrypt,
+)
+
+
+Input = TypeVar("Input")
+Output = TypeVar("Output")
+
+
+class AwsCryptographicMaterialProviders:
+    """//////////
```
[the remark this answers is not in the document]
lucasmcdonald3: See [explanation](https://github.com/smithy-lang/smithy-dafny/blob/865dd7522c9ff7c708f58f4d305f253a3c31ef75/codegen/smithy-dafny-codegen/src/main/java/software/amazon/polymorph/smithypython/localservice/extensions/DafnyPythonLocalServiceClientCodegenPlugin.java#L102-L105)
Though it seems that logic misses this case...
I'll look more at this

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/plugin.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def set_config_impl(config: Config):
+    """
+    Set the Dafny-compiled implementation in the Smithy-Python client Config
+    and load our custom NoRetriesStrategy.
+    """
+    config.dafnyImplInterface = DafnyImplInterface()
+    if isinstance(config, KeyStoreConfig):
+        from software_amazon_cryptography_keystore_internaldafny import default__
```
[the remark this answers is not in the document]
lucasmcdonald3: It is wild since it is an anti-pattern: https://peps.python.org/pep-0008/#imports
This is done to avoid circular dependencies when importing generated Dafny code

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/models.py
```diff
@@ -0,0 +1,944 @@
...
+        return {}
+
+    @staticmethod
+    def from_dict(d: Dict[str, Any]) -> "CreateKeyStoreInput":
+        """Creates a CreateKeyStoreInput from a dictionary.
+
+        The dictionary is expected to use the modeled shape names rather than the
+        parameter names as keys to be mostly compatible with boto3.
+        """
+        return CreateKeyStoreInput()
```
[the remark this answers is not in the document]
lucasmcdonald3: This is actually correct, CreateKeystoreInput is an empty structure. [link](https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/Model/KeyStore.smithy#L139-L140)
There's an argument that this method shouldn't exist on models that don't have any members, but I don't know if this is worth fixing

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/models.py
```diff
@@ -0,0 +1,944 @@
...
+
+    @staticmethod
+    def from_dict(d: Dict[str, Any]) -> "VersionKeyOutput":
+        """Creates a VersionKeyOutput from a dictionary.
+
+        The dictionary is expected to use the modeled shape names rather than the
+        parameter names as keys to be mostly compatible with boto3.
+        """
+        return VersionKeyOutput()
+
```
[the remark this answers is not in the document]
lucasmcdonald3: Ditto to above, this is an empty structure. [link](https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/Model/KeyStore.smithy#L201-L202).
This is mostly a correctness thing.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/errors.py
```diff
@@ -0,0 +1,193 @@
...
+            "message": d["message"],
+        }
+
+        return KeyStoreException(**kwargs)
+
+    def __repr__(self) -> str:
+        result = "KeyStoreException("
+        result += f"message={self.message},"
+        if self.message is not None:
+            result += f"message={repr(self.message)}"
```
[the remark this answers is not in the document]
lucasmcdonald3: Yes, it does. I'll look into this

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/deserialize.py
```diff
@@ -0,0 +1,122 @@
...
+    _sdk_error_to_dafny_error as com_amazonaws_dynamodb_sdk_error_to_dafny_error,
+)
+from com_amazonaws_kms.smithygenerated.com_amazonaws_kms.shim import (
+    _sdk_error_to_dafny_error as com_amazonaws_kms_sdk_error_to_dafny_error,
+)
+
+from .config import Config
+
+
+async def _deserialize_get_key_store_info(input: DafnyResponse, config: Config):
```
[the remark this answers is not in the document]
lucasmcdonald3: I'm not certain actually... I wonder if this is used somewhere that I don't remember of the top of my head, or if this was used at some point but isn't used anymore...
I can look into this if it's blocking but would plan to leave it as-is otherwise

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/dafny_to_smithy.py
```diff
@@ -0,0 +1,215 @@
...
+)
+
+
+def smithy_api_Unit():
+    return (
+        aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_keystore.models.Unit()
+    )
+
+
+def aws_cryptography_keystore_CreateKeyStoreInput(input):
```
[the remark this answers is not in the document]
lucasmcdonald3: Yes, I need to fix this. Will fix.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/dafny_to_smithy.py
```diff
@@ -0,0 +1,215 @@
...
+    return aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_keystore.models.CreateKeyInput(
+        branch_key_identifier=(
+            (input.branchKeyIdentifier.value.VerbatimString(False))
+            if (input.branchKeyIdentifier.is_Some)
+            else None
+        ),
+        encryption_context=(
+            (
+                {
+                    bytes(key): bytes(value)
```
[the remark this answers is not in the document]
lucasmcdonald3: Yes, we should do this.
I've already done this in the PR that runs against CI: [link](https://github.com/aws/aws-cryptographic-material-providers-library/blob/c8b87f2afd4f24aa13db9f72e1e9695bd9e9e12a/AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/dafny_to_smithy.py#L33-L39)
I'll merge this into this PR.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/serialize.py
```diff
@@ -0,0 +1,65 @@
+# Code generated by smithy-python-codegen DO NOT EDIT.
+
+import aws_cryptographic_materialproviders.smithygenerated.aws_cryptography_keystore.smithy_to_dafny
+
+from .dafny_protocol import DafnyRequest
+
+from .config import Config
+
+
+async def _serialize_get_key_store_info(input, config: Config) -> DafnyRequest:
```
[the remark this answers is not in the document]
lucasmcdonald3: https://github.com/aws/aws-cryptographic-material-providers-library/pull/225/#discussion_r1546845872

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/errors.py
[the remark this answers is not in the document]
lucasmcdonald3: This is some garbage from Smithy-Python...
I'll try to make it pluggable and overwrite it

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/client.py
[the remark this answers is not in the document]
lucasmcdonald3: Oh geez, this might be more involved. It seems that `ServiceShapes` have special properties when it comes to DocumentationTraits...

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/errors.py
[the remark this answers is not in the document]
lucasmcdonald3: Fixed

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/dafny_to_smithy.py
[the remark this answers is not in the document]
lucasmcdonald3: Changed this file to `dafny_input`; `smithy_to_dafny.py` uses `native_input`

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/client.py
```diff
@@ -0,0 +1,462 @@
+# Code generated by smithy-python-codegen DO NOT EDIT.
+
+import asyncio
+from asyncio import sleep
+from copy import deepcopy
```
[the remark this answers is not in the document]
lucasmcdonald3: Removed

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_keystore/plugin.py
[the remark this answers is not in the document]
justplaz: ah yeah i might have had a similar comment on a previous PR...

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/smithygenerated/aws_cryptography_materialproviders/client.py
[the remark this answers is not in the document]
justplaz: that's fine...weird, but fine.

justplaz (approved):
LGTM, remaining stuff isn't blocking.

PR #230, opened 2024-03-02
chore: Use shared SmithyDafnyMakefile.mk

robin-aws:
_Description of changes:_

Replaces nearly all of `SharedMakefileV2.mk` with the new `SmithyDafnyMakefile.mk` from the `smithy-dafny` submodule. What's left is just the repo-specific configuration of variables.

See https://github.com/aws/aws-cryptographic-material-providers-library/pull/233/ for a more readable diff of the existing 
`SharedMakefileV2.mk` with `SmithyDafnyMakefile.mk`, showing what targets are changing (which is very little as the extra target/improvements in the former were mostly carried over to the latter already).


seebees (approved):
Let it begin!

PR #252, opened 2024-03-27
chore: add useful utf8 lemmas

ajewellamz:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/StandardLibrary/src/UTF8.dfy
+++ b/StandardLibrary/src/UTF8.dfy
@@ -54,7 +54,7 @@ module {:extern "UTF8"} UTF8 {
     Failure(error)
   }
 
- // must be true for any correct UTF8 implementation
+  // must be true for any correct UTF8 implementation
   lemma {:axiom} Utf8EncodeUnique(x : string, y : string)
     requires Encode(x).Success?
     requires Encode(y).Success?
--- a/StandardLibrary/src/UTF8.dfy
+++ b/StandardLibrary/src/UTF8.dfy
@@ -54,6 +54,36 @@ module {:extern "UTF8"} UTF8 {
     Failure(error)
   }
 
+ // must be true for any correct UTF8 implementation
+  lemma {:axiom} Utf8EncodeUnique(x : string, y : string)
+    requires Encode(x).Success?
+    requires Encode(y).Success?
+    ensures x == y ==> Encode(x).value == Encode(y).value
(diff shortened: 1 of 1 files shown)


StandardLibrary/src/UTF8.dfy
```diff
@@ -54,6 +54,36 @@ module {:extern "UTF8"} UTF8 {
     Failure(error)
   }
 
+  // must be true for any correct UTF8 implementation
+  lemma {:axiom} Utf8EncodeUnique(x : string, y : string)
+    requires Encode(x).Success?
+    requires Encode(y).Success?
+    ensures x == y ==> Encode(x).value == Encode(y).value
+    ensures x != y ==> Encode(x).value != Encode(y).value
```
seebees, line 62: `Encode` is a function, so you don't need this.

StandardLibrary/src/UTF8.dfy
```diff
@@ -54,6 +54,36 @@ module {:extern "UTF8"} UTF8 {
...
+    ensures !(x <  y) ==> !(Encode(x).value <  Encode(y).value)
+    ensures !(y <= x) ==> !(Encode(y).value <= Encode(x).value)
+    ensures !(y <  x) ==> !(Encode(y).value <  Encode(x).value)
+
+  // must be true for any correct UTF8 implementation
+  lemma {:axiom} Utf8DecodeUnique(x : seq<uint8>, y : seq<uint8>)
+    requires Decode(x).Success?
+    requires Decode(y).Success?
+    ensures x == y ==> Decode(x).value == Decode(y).value
+    ensures x != y ==> Decode(x).value != Decode(y).value
```
seebees, line 77: `Decode` is a function, so you don't need this.

StandardLibrary/src/UTF8.dfy
seebees, line 1: I think that the thing you want is that `Encode` and `Decode` are distributive.

Like this. I'm not endorsing this, but I think you could use this to prove all your claims right?

```dafny
lemma {:axiom} Utf8EncodeIsDistributive(x: string)
    requires Encode(x).Success?
    ensures forall head, tail | head + tail == x
      :: Encode(head).Success?
          ==>
            && Encode(tail).Success?
            && Encode(head).value + Encode(tail).value == Encode(x).value
```

PR #255, opened 2024-03-28
chore(MPL): Migrate release process from CB to GHA CodeBuild Runners

josecorella:
_Issue #, if available:_

_Description of changes:_


.github/workflows/java-release.yml
```diff
@@ -8,13 +8,196 @@ on:
...
+
+      - name: Retrieve Keys from SecretsManager and add keys to KeyChain
+        run: |
+          sudo apt-get install gnupg2 -y
+          aws secretsmanager get-secret-value --region us-west-2 --secret-id Maven-GPG-Keys-Release --query SecretBinary --output text | base64 -d > ~/mvn_gpg.tgz
+          tar -xvf ~/mvn_gpg.tgz -C ~
+          mkdir /root/.gnupg
+          printf "use-agent\npinentry-mode loopback" >> ~/mvn_gpg/gpg.conf
+          printf "allow-loopback-pinentry" >> ~/mvn_gpg/gpg-agent.conf
+          mv -v ~/mvn_gpg/* /root/.gnupg/
```
texastony, line 69: Question: When this is run, who can see the GHW logs?
Can we restrict to just "us"?

.github/workflows/java-release.yml
```diff
@@ -8,13 +8,196 @@ on:
...
+    strategy:
+      matrix:
+        library: [AwsCryptographicMaterialProviders]
+        dotnet-version: ["6.0.x"]
+    runs-on: codebuild-AWS-MPL-Release-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-5.0-large
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      REGION: us-east-1
```
texastony, line 42: Question: If the host is in `us-east-1`, do all our tests know to call `us-west-2` for the test resources?
We may need to override this variable when running tests.

.github/workflows/java-release.yml
```diff
@@ -8,13 +8,196 @@ on:
...
+          make transpile_test_java
+          # force gradle to ignore cache and rebuild dependencies; guarantess we are using the published artifact
+          gradle -p runtimes/java build --refresh-dependencies
+
+          # assume role manually 
+          TMP_ROLE=$(aws sts assume-role --role-arn "arn:aws:iam::370957321024:role/GitHub-CI-MPL-Dafny-Role-us-west-2" --role-session-name "CB-TestVectorResources")
+          export TMP_ROLE
+          export AWS_ACCESS_KEY_ID=$(echo "${TMP_ROLE}" | jq -r '.Credentials.AccessKeyId')
+          export AWS_SECRET_ACCESS_KEY=$(echo "${TMP_ROLE}" | jq -r '.Credentials.SecretAccessKey')
+          export AWS_SESSION_TOKEN=$(echo "${TMP_ROLE}" | jq -r '.Credentials.SessionToken')
```
texastony, line 189: Suggestion: Again, override the AWS_REGION variable to be us-west-2 for tests to pass.

.github/workflows/net-lib-release.yml
```diff
@@ -0,0 +1,395 @@
...
+      - name: make ${{inputs.project}}
+        shell: bash
+        working-directory: ./${{inputs.project}}
+        run: |
+          make transpile_net
+          make test_net FRAMEWORK=net6.0
+      - name: Unset AWS Creds
+        shell: bash
+        run: |
+          echo "AWS_ACCESS_KEY_ID=" >> $GITHUB_ENV
```
texastony, line 129: Question: `sh` has the `unset` command.
Does GitHub env not have an allegory?

cfn/CI.yaml
```diff
@@ -53,6 +53,7 @@ Resources:
                     "arn:aws:iam::587316601012:role/service-role/codebuild-AWS-MPL-Java-service-role-release",
                     "arn:aws:iam::587316601012:role/service-role/codebuild-AWS-MPL-DotNet-service-role-release",
                     "arn:aws:iam::587316601012:role/service-role/codebuild-AWS-ESDK-DotNet-service-role-release",
+                    "arn:aws:iam::587316601012:role/service-role/codebuild-AWS-MPL-service-role-release",
```
texastony, line 56: Blocking/Suggestion: This cannot be given to the CI Role or any PR CI can assume it.
We must create a new GitHub Role explicitly for releasing and assume that.
```suggestion
```

I know that we have other protections to prevent this abuse,
but least privileged strongly suggests that we keep this role/CFN template CI oriented.

Create a different IAM Role explicitly for releasing.
Do not allow `ToolsDevelopment` or any other CI role to assume it.

texastony (changes requested):
Only Blocking issue is giving the CI role release secretes.
Create a distinct role for releasing,
and allow GitHub to assume that ONLY if the branch is `main`.

Ideally, we would put additional constraints on the assumption of that role,
if you can think of others.

.github/workflows/java-release.yml
josecorella: yes, all tests know to call us-west-2. This is not new, all CB Release builds are hosted in us-east-1 calling us-west-2 resources

cfn/CI.yaml
josecorella: I am confused. AFAIK this doesn't do that. This gives the codebuild service role permissions to assume the CI role used to test, It DOES NOT let the CI role assume the codebuild service role.

texastony: Oh you are right!
I am an idiot.

.github/workflows/net-lib-release.yml
josecorella: yes, but in gha land `unset` will only work for that step in the job. It won't carry over to the next step.

.github/workflows/java-release.yml
josecorella: I don't think you can restrict who can see the GHW log, but why would we? Do we care? All secrets are abstracted in the log so GH will not let you log them.
In either case, for this step we are not printing any creds or anything crazy.

josecorella:
force push due to rebase

PR #262, opened 2024-04-02
chore: add example of tail recursion

ajewellamz:
_Issue #, if available:_

_Description of changes:_


TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/ParseJsonManifests.dfy
```diff
@@ -41,8 +41,23 @@ module {:options "-functionSyntax:4"} ParseJsonManifests {
   }
   by method {
     // This function ideally would be`{:tailrecursion}`
-    // but it is not simple to here is a method
-    // so that it does not explode with huge numbers of tests.
+    // immediately below is an example of how to make it tail recursive
+    // However, we're leaving this as "by method" to provide an example of how to do that
```
seebees, line 45:
```suggestion
    // at the time this did not seem simple.
    // Here is an example of how to make it tail recursive that may work.
    // However, we're leaving this as "by method"
    // to avoid changing anything and to provide an example of how to do that
```

seebees (commented):
What about that?

seebees (approved):
LGTM

PR #266, opened 2024-04-03
Python MPL-Core: Dafny extern

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_

Add StormTrackingCMC extern implementation.
This will only be used in single-threaded environments for MPL launch. (Only used in HKeyring, which is labelled not-thread-safe for launch.)

NET impl: https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/runtimes/net/Extern/StormTrackingCMC.cs
Java impl: https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/runtimes/java/src/main/java/software/amazon/cryptography/internaldafny/StormTrackingCMC/StormTrackingCMC.java


seebees (commented):
This looks correct to me. I would like to get a more experienced Python person to validate my eyes.

texastony:
@lucasmcdonald3 we may want to rely on Dafny's implementation:
https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std_Concurrent.py

lucasmcdonald3:
> @lucasmcdonald3 we may want to rely on Dafny's implementation: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std_Concurrent.py

I would like to push this out of scope for now --
* The existing implementation matches the behavior for Java/NET
* We need extra work to pull this in from Dafny
I would suggest that "migrating cache extern implementation to use Dafny implementation" should be done at the same time for Java, NET, and Python

texastony:
> > @lucasmcdonald3 we may want to rely on Dafny's implementation: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std_Concurrent.py
> 
> I would like to push this out of scope for now -- ...

I was really suggesting we just copy and paste,
rather than formally depend.

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/internaldafny/extern/SynchronizedLocalCMC.py
```diff
@@ -0,0 +1,70 @@
...
+        return val
+
+    def DeleteCacheEntry_k(self, input):
+        self.lock.Lock__()
+        val = self.wrapped.DeleteCacheEntry(input)
+        self.lock.Unlock()
+        return val
+    
+    def __str__(self):
+        return "LocalCMC.StormTrackerCMC"
```
josecorella, line 68: shouldn't this be
```suggestion
        return "LocalCMC.SynchronizedCMC"
```
?

josecorella (changes requested):

AwsCryptographicMaterialProviders/runtimes/python/src/aws_cryptographic_materialproviders/internaldafny/extern/SynchronizedLocalCMC.py
lucasmcdonald3: Good catch, 1 sec

```suggestion
        return "LocalCMC.SynchronizedLocalCMC"
```

lucasmcdonald3: Mirroring https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/runtimes/java/src/main/java/software/amazon/cryptography/internaldafny/SynchronizedLocalCMC/SynchronizedLocalCMC.java#L90

josecorella (approved):
lgtm

PR #296, opened 2024-04-25
chore: add test for VersionKey with an MRK

ajewellamz:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/.github/workflows/library_dafny_verification.yml
+++ b/.github/workflows/library_dafny_verification.yml
@@ -28,7 +28,7 @@ jobs:
             AwsCryptographicMaterialProviders,
             StandardLibrary,
           ]
-        os: [macos-12-large]
+        os: [macos-latest-large]
     runs-on: ${{ matrix.os }}
     defaults:
       run:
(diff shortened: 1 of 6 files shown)


texastony (approved):

.github/workflows/library_dafny_verification.yml
```diff
@@ -28,7 +28,7 @@ jobs:
             AwsCryptographicMaterialProviders,
             StandardLibrary,
           ]
-        os: [macos-latest-large]
+        os: [macos-12-large]
```
texastony, line 31: We should leave this one alone.
It does not run any code, just verifies it.
```suggestion
        os: [macos-latest-large]
```

texastony (approved):
Approve with Suggestion.

josecorella (approved):

texastony (approved):
To prove that the persisted KMS ARN is not changed by MRK Version,
we would need to compare the KMS ARN element of the BranchKeyMaterails' Encryption Context in `oldActiveResult` and `newActiveResult`.

Thus, this new Test really only asserts that Version succeeded and is use-able.
It does not assert if the KMS ARN has been mutated.

PR #306, opened 2024-05-01
ci: Cut Dafny issues on nightly failures

robin-aws:
_Description of changes:_

Automates cutting an issue to `dafny-lang/dafny` if the nightly build against the latest Dafny prerelease fails.

Reuses the same GH token as the semantic release workflows, since that already has more than enough permissions (only `public_repo` is needed for this action).

Tested this by putting the same steps in the manual CI workflow, which created this issue: https://github.com/dafny-lang/dafny/issues/5391

### DIFF
--- a/.github/workflows/nightly_dafny.yml
+++ b/.github/workflows/nightly_dafny.yml
@@ -40,6 +40,9 @@ jobs:
 
   cut-issue-on-failure:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     needs:
       [
         dafny-nightly-format,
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_dafny.yml
```diff
@@ -37,3 +37,43 @@ jobs:
...
+      ]
+    if: ${{ always() && contains(needs.*.result, 'failure') }}
+    steps:
+      # We need access to the role that is able to get CI Bot Creds
+      - name: Configure AWS Credentials for Release
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: us-west-2
+          role-to-assume: arn:aws:iam::587316601012:role/GitHub-CI-CI-Bot-Credential-Access-Role-us-west-2
+          role-session-name: CI_Bot_Release
```
josecorella, line 61: small nit:

```suggestion
          role-session-name: Dafny_Issue_Blocker
```

PR #436, opened 2024-06-20
chore: Add more robust public key validation tests

josecorella:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/AwsCryptographyPrimitives/runtimes/java/src/main/java/ECDH/ECCUtils.java
+++ b/AwsCryptographyPrimitives/runtimes/java/src/main/java/ECDH/ECCUtils.java
@@ -3,12 +3,14 @@ package ECDH;
 import static ECDH.ECCAlgorithm.eccAlgorithm;
 import static software.amazon.smithy.dafny.conversion.ToDafny.Simple.ByteSequence;
 
+import Random_Compile.ExternRandom;
 import Signature.PublicKeyUtils;
 import Signature.PublicKeyUtils;
 import Signature.SignatureAlgorithm;
 import StandardLibraryInternal.InternalResult;
 import StandardLibraryInternal.InternalResult;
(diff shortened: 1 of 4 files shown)


AwsCryptographyPrimitives/runtimes/java/src/main/java/ECDH/ECCUtils.java
```diff
@@ -378,4 +406,109 @@ public static java.security.spec.ECParameterSpec ecParameterSpec(
       java.security.spec.ECParameterSpec.class
     );
   }
+
+  public static Result<
```
justplaz, line 410: nit: if this is mostly for testing, and always fails, we should probably include a javadoc comment saying so, so that consumers don't mistake this for a useful utility function...it also could maybe be moved into a separate class, but i don't feel so strongly about it.

AwsCryptographyPrimitives/runtimes/java/src/main/java/ECDH/ECCUtils.java
```diff
@@ -378,4 +406,109 @@ public static java.security.spec.ECParameterSpec ecParameterSpec(
...
+          AwsCryptographicPrimitivesError
+            .builder()
+            .message(e.getMessage())
+            .build()
+        )
+      );
+    }
+  }
+
+  public static Result<
```
justplaz, line 459: ditto

AwsCryptographyPrimitives/runtimes/java/src/main/java/ECDH/ECCUtils.java
josecorella: ack; I can add a small comment. This extern is not exposed in the primitives client, so if someone were to use this they are breaking the primitive client contract anyways...

justplaz (approved):

PR #488, opened 2024-07-08
chore: bump dafnyVerifyVersion to 4.7.0

josecorella:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/project.properties
+++ b/project.properties
@@ -6,4 +6,4 @@
 # The Java project include this file as a Gradle properties
 # And the Dotnet projects include and parse this file.
 dafnyVersion=4.2.0
-dafnyVerifyVersion=4.2.0
+dafnyVerifyVersion=4.7.0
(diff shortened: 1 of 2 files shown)


texastony (commented):
@josecorella You may want to check with @robin-aws , but maybe we should add new codegen patches for 4.7?

josecorella:
I will, this is baby steps though

josecorella:
I clicked on create instead of draft

project.properties
```diff
@@ -6,4 +6,4 @@
 # The Java project include this file as a Gradle properties
 # And the Dotnet projects include and parse this file.
 dafnyVersion=4.2.0
-dafnyVerifyVersion=4.2.0
+dafnyVerifyVersion=4.7.0
```
robin-aws, line 9: Just FYI, I've had to add support to smithy-dafny to generate these project.properties files in order for the nightly builds to be able to use a newer Dafny everywhere. It isn't yet outputting this property, so that means the nightly builds are no longer verifying with the latest Dafny master.

The good news is this change helps close that gap, but it still exists and we should track it somehow.

robin-aws:
> You may want to check with @robin-aws , but maybe we should add new codegen patches for 4.7?

No need, this is why the codegen patches apply for any Dafny version at or above the version in their name.

PR #508, opened 2024-07-11
chore: fix mpl java test vector project version and bump

josecorella:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
+++ b/TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
@@ -19,7 +19,7 @@ var props = Properties().apply {
 var dafnyVersion = props.getProperty("dafnyVersion")
 
 group = "software.amazon.cryptography"
-version = "1.5.1-SNAPSHOT"
+version = "1.0-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


TestVectorsAwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
```diff
@@ -19,7 +19,7 @@ var props = Properties().apply {
 var dafnyVersion = props.getProperty("dafnyVersion")
 
 group = "software.amazon.cryptography"
-version = "1.5.1-SNAPSHOT"
+version = "1.0-SNAPSHOT"
```
rishav-karanjit, line 22: I am confused. Why is this version 1.0?

josecorella: the test vectors project is unpublished.
It's version used to be 1.0-SNAPSHOT but an update to the releaserc.cjs file bumped this as well, it should not have.

The DBESDK and ESDK depend on this project as 1.0-SNAPSHOT  not 1.5.1-SNAPSHOT as a result if we try to build the latest test vectors in the dbesdk it fails because it is looking for 1.0-SNAPSHOT

rishav-karanjit (approved):

PR #517, opened 2024-07-15
chore(verfication): remove {:only}

josecorella:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/.github/workflows/check_only_keyword.yml
+++ b/.github/workflows/check_only_keyword.yml
@@ -24,7 +24,7 @@ jobs:
          # checking in code with the dafny decorator {:only}
          # will not verify the entire file or maybe the entire project depending on its configuration
          # This action checks if you are either adding or removing the {:only} decorator and posting on the pr if you are. 
-          echo "ONLY_KEYWORD=$(git diff origin/main origin/${GITHUB_HEAD_REF} **/*.dfy | grep -i {:only}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


.github/workflows/check_only_keyword.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Check if ONLY_KEYWORD is not empty
+        id: comment
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ONLY_KEYWORD: ${{ steps.only-keyword.outputs.ONLY_KEYWORD }}
+        if: ${{env.ONLY_KEYWORD != ''}}
+        run: |
+          COMMENT="It looks like you are adding or removing the dafny keyword {:only}.\nIs this intended?"
```
seebees, line 37: Why not just have this action fail? If I'm working on a PR and it does not pass for this reason, that feels like a good thing?

seebees (changes requested):

seebees (approved):
LGTM

josecorella:
Although the action, `Check {:only} decorator presence / grep-only-verification-keyword` is failing it is safe to merge in this instance since we are removing it. The action we have set up is very basic and is made to ensure we fail if we either add it or remove it from code. Since this is the only occurrence of `{:only}` in the code, the action will never get triggered for removing it, only if we add it. I will merge this as is and make it a required action in the branch protection settings.

PR #537, opened 2024-07-25
Python: Testvectors Externs

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/runtimes/python/test/internaldafny/extern/GetTestDirectoryExtern.py
+++ b/TestVectorsAwsCryptographicMaterialProviders/runtimes/python/test/internaldafny/extern/GetTestDirectoryExtern.py
@@ -1,0 +1,8 @@
+import TestWrappedMaterialProvidersMain
+import os
+import _dafny
+
+def GetTestVectorExecutionDirectory():
+    return _dafny.Seq(os.getcwd() + "/../../")
+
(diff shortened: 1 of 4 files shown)


TestVectorsAwsCryptographicMaterialProviders/runtimes/python/test/internaldafny/extern/GetTestDirectoryExtern.py
```diff
@@ -0,0 +1,8 @@
+import TestWrappedMaterialProvidersMain
+import os
+import _dafny
+
+def GetTestVectorExecutionDirectory():
+    return _dafny.Seq(os.getcwd() + "/../../")
```
josecorella, line 6: If we were to publish the python test vectors how would this behave? Is it always the case that the execrution dir has to be two dirs up?

lucasmcdonald3: This is only used in the Dafny tests.
We only really run these via `make test_python`, which enforces the execution directory as `runtimes/python`, which means the `/../../` is valid for any tests run from `make test_python`.
(This is also true for Java/NET tests, those also have an implicit expected root directory of `""`)

josecorella: that's class thank you!

josecorella (approved):

PR #563, opened 2024-08-01
chore: do not use hash as a variable or member name, because Rust gets confused

ajewellamz:
_Issue #, if available:_

_Description of changes:_


lucasmcdonald3 (commented):
Can the Dafny-Rust compiler escape keywords? ex. `hash` -> `_hash` or `hash_`
If so let's just add `hash` to the escaped keywords.

PR #575, opened 2024-08-07
fix: GetCurrentTimeStamp returns ISO8601 format

seebees:
The Java extern returned `YYYY-MM-DDTHH:mm:ss:ssssssZ` This is not ISO8601.

Also, time is non-deterministic,
therefore GetCurrentTimeStamp MUST be a method.


josecorella (approved):
nice catch!

PR #589, opened 2024-08-12
chore: Rename AtomicPrimitives Dafny module name

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_

Change the Dafny name for this client to AtomicPrimitives.

Both Go and Python will reference every Local Service client module name as the localService trait's sdkId value.
The client module name is the same as the sdkId in almost all cases, except Primitives.
This changes Primitives' Dafny name to align with its localService trait sdkId value.
(There is no change for Java/NET compiled code here because this module is an extern in those languages.)

Downstream builds:
This verifies that with ESDK/DBESDK can build with these changes. (Downstream libraries need to make similar changes, but we can't merge these into the respective main branches yet, and we should only pin those libraries to released versions of the MPL.)
* ESDK: https://github.com/aws/aws-encryption-sdk-dafny/pull/673
* DBESDK: https://github.com/aws/aws-database-encryption-sdk-dynamodb/pull/1299


ajewellamz (dismissed):
LGTM

AwsCryptographyPrimitives/src/Index.dfy
texastony, line 1:
```
cd /workplace/aws-database-encryption-sdk-dynamodb-java/DynamoDbEncryption
grep "Aws.Cryptography.Primitives" runtimes/java/src/main/dafny-generated/**/*.java
# no hits
grep "software.amazon.cryptography.primitives" runtimes/java/src/main/dafny-generated/**/*.java
# many hits, but they all reflect the "extern" name, not the Modules name, so this non-breaking.
```

ajewellamz (approved):

PR #618, opened 2024-08-21
fix: Remove uses of `:|`

atomb:
_Issue #, if available:_

_Description of changes:_

Removes uses of `:|` in Dafny code, and updates the libraries submodule to a version that does not use it, either. This will eliminate failures when compiling with Dafny 4.8.0 and `-definiteAssignment:3`.

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMMs/DefaultCMM.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMMs/DefaultCMM.dfy
@@ -330,7 +330,7 @@ module DefaultCMM {
     //# - The operations made on the encryption context on the Get Encryption Materials call
     //# SHOULD be inverted on the Decrypt Materials call.
 
-    method DecryptMaterials'(
+    method {:vcs_split_on_every_assert} DecryptMaterials'(
       input: Types.DecryptMaterialsInput
     )
       returns (output: Result<Types.DecryptMaterialsOutput, Types.Error>)
@@ -373,6 +373,7 @@ module DefaultCMM {
       //# the values MUST be equal or the operation MUST fail.
       ensures
         && (output.Success? ==> CMM.ReproducedEncryptionContext?(input))
+      ensures
         && (!CMM.ReproducedEncryptionContext?(input) ==> output.Failure?)
       //= aws-encryption-sdk-specification/framework/cmm-interface.md#decrypt-materials
       //= type=implication
@@ -479,16 +480,20 @@ module DefaultCMM {
       var requiredEncryptionContextKeys := [];
       if input.reproducedEncryptionContext.Some? {
         var keysSet := input.reproducedEncryptionContext.value.Keys;
+        ghost var keysSet' := keysSet;
         var keysSeq := SortedSets.ComputeSetToSequence(keysSet);
         var i := 0;
         while i < |keysSeq|
+          invariant Seq.HasNoDuplicates(keysSeq)
+          invariant forall j | 0 <= j < i && i < |keysSeq| :: keysSeq[j] !in keysSet'
+          invariant forall j | i <= j < |keysSeq| :: keysSeq[j] in keysSet'
+          invariant |keysSet'| == |keysSeq| - i
           invariant forall key
                       |
-                      && multiset(keysSeq) == multiset(keysSet)
-                      && key in input.reproducedEncryptionContext.value
+                      && key in input.reproducedEncryptionContext.value.Keys
                       && key in input.encryptionContext
-                      && key !in keysSet
-                      :: input.reproducedEncryptionContext.value[key] == input.encryptionContext[key]
+                      && key !in keysSet'
+                      :: input.encryptionContext[key] == input.reproducedEncryptionContext.value[key]
           invariant forall key <- requiredEncryptionContextKeys
                       :: key !in input.encryptionContext
         {
@@ -500,7 +505,11 @@ module DefaultCMM {
           } else {
             requiredEncryptionContextKeys :=  requiredEncryptionContextKeys + [key];
           }
+          keysSet' := keysSet' - {key};
           i := i + 1;
+          assert forall j | i <= j < |keysSeq| :: keysSeq[j] in keysSet' by {
+            reveal Seq.HasNoDuplicates();
+          }
         }
       }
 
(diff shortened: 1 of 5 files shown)


texastony:
@atomb I put this PR in draft. Once verification and other CI is passing, you can mark it ready for review.

StandardLibrary/src/StandardLibrary.dfy
```diff
@@ -335,7 +335,7 @@ module StandardLibrary {
    * The function is compilable, but will not exhibit enviable performance.
    */
 
-  function method {:tailrecursion} SetToOrderedSequence<T(!new,==)>(s: set<seq<T>>, less: (T, T) -> bool): (q: seq<seq<T>>)
+  function SetToOrderedSequence<T(!new,==)>(s: set<seq<T>>, less: (T, T) -> bool): (q: seq<seq<T>>)
```
robin-aws, line 338: This will be breaking for consumers of the MPL, since you're changing it from compiled to ghost, but there are compiled references to it already, e.g. https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/DynamoDbEncryption/dafny/DynamoDbEncryption/src/DynamoToStruct.dfy#L1091

Are you relying on assuming we can replace all such references with `ComputeSetToSequence[2]`?

atomb: Ah, I removed internal compiled references to it but didn't realize it was part of the exported API.

I suspect we should be able to replace them all, though it's possible there are complexities I'm not thinking of.

atomb: I think the two instances in that file are the only place it's used in a compiled context, and I think it can be replaced by the `Compute` version there.

josecorella (dismissed):
starred long and hard at the change and believe this is a fine change.

StandardLibrary/src/StandardLibrary.dfy
texastony: While we can logically replace them, we cannot assume that the code on our customers computer will not be updated.

i.e: Our customers will still have references to the old version, and rely on the references being provided by the MPL.

Unless both of the references in the DB-ESDK, and any references in the ESDK, are only in ghost code and have ALWAYS been in only in ghost code.

texastony (changes requested):
Blocking until breaking change issue is resolved.

StandardLibrary/src/StandardLibrary.dfy
robin-aws: @atomb Could you make this a `function by method` instead and use the `ComputeSetToSequence2` implementation in the by method body?

robin-aws: Actually I don't think you have to change this anyway, I don't get an error when I do `dafny verify src/StandardLibrary.dfy --function-syntax:3 --enforce-determinism`.

Since this `:|` is in an expression context it already has to be provably deterministic.

atomb: Oh, you are indeed correct. I've changed it back.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMMs/DefaultCMM.dfy
```diff
@@ -478,25 +480,36 @@ module DefaultCMM {
...
+          invariant |keysSet'| == |keysSeq| - i
           invariant forall key
                       |
-                      && key in input.reproducedEncryptionContext.value
+                      && key in input.reproducedEncryptionContext.value.Keys
                       && key in input.encryptionContext
-                      && key !in keysSet
-                      :: input.reproducedEncryptionContext.value[key] == input.encryptionContext[key]
+                      && key !in keysSet'
+                      :: input.encryptionContext[key] == input.reproducedEncryptionContext.value[key]
```
seebees, line 496: Slight nit, adjusting the verification makes it a little hard to verify the lack of change :)
This way would have made it clear that the only actual change was to change the invariant to use the ghost value.
and that everything else was additions.

 ```dafny
                      && key in input.reproducedEncryptionContext.value
                      && key in input.encryptionContext
                      && key !in keysSet'
                      :: input.reproducedEncryptionContext.value[key] == input.encryptionContext[key]
```

File: AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMMs/RequiredEncryptionContextCMM.dfy [hunk 3]
```diff
@@ -52,16 +53,7 @@ module RequiredEncryptionContextCMM {
...
-      while keySet != {}
-        invariant |keySeq| + |keySet| == |inputKeys|
-        invariant forall k <- keySeq
-                    :: k in inputKeys
-      {
-        var key :| key in keySet;
-        keySeq := keySeq + [key];
-        keySet := keySet - {key};
-      }
+      var keySeq := SortedSets.ComputeSetToSequence(keySet);
```
seebees, line 56: Similarly here, I think that we could see the diff more clearly as invariant by writing it this way.
(I checked the ensures clause)
This way the `invariant` changes to an assert.

```dafny
var keySeq := SortedSets.ComputeSetToSequence(inputKeys);
assert |keySeq| == |inputKeys|;
assert forall k <- keySeq :: k in inputKeys;
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CanonicalEncryptionContext.dfy
```diff
@@ -25,7 +26,7 @@ module CanonicalEncryptionContext {
   {
     :- Need(|encryptionContext| < UINT16_LIMIT,
             Types.AwsCryptographicMaterialProvidersException( message := "Encryption Context is too large" ));
-    var keys := SetToOrderedSequence(encryptionContext.Keys, UInt.UInt8Less);
+    var keys := SortedSets.ComputeSetToOrderedSequence2(encryptionContext.Keys, UInt.UInt8Less);
```
seebees, line 29: Looking at `SortedSets.ComputeSetToOrderedSequence2` I can see `ensures res == SetToOrderedSequence(s, less)` so this MUST be the same.

AwsCryptographyPrimitives/test/TestRSA.dfy
seebees, line 1: These testing changes look fine.

seebees (approved):
These changes look good.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMMs/DefaultCMM.dfy
atomb: Good point. Those changes are an artifact of attempting to reduce brittleness, before I realized that it was actually missing information leading to failure here. They're not necessary, and I'd be happy to undo them.

seebees (approved):
Wonderful!

PR #631, opened 2024-08-23
feat(KeyStoreAdmin):

texastony:
_Issue #, if available:_

_Description of changes:_


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/Model/KeyStoreAdmin.smithy
```diff
@@ -0,0 +1,152 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+namespace aws.cryptography.keyStoreAdmin
```
texastony, line 3: Maybe the namespace should just be keyStore.

Can more than one Service be a in a Namespace?

texastony: Oh, no, that would mess things up.
Because then there would be two VersionKeyInputs in the same namespace...

PR #705, opened 2024-09-11
fix: branch key version MUST be 16 bytes

seebees:
This is maintained by the key store
on branch key creation.
But if branch keys are created
outside of the library,
they may not match the hierarchical keyrings needs.

Specifically the serialization of the encrypted data key requires a fixed for the version.

And a spec update


AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsHierarchicalKeyring.dfy
```diff
@@ -253,6 +253,16 @@ module AwsKmsHierarchicalKeyring {
...
       var branchKeyVersion := hierarchicalMaterials.branchKeyVersion;
       var branchKeyVersionAsString :- UTF8.Decode(branchKeyVersion).MapFailure(WrapStringToError);
+      //= aws-encryption-sdk-specification/framework/aws-kms/aws-kms-hierarchical-keyring.md#ciphertext
+      //#concatenated with the byte representation of the UUID branch key version from the [branch key materials](../structures.md#branch-key-materials)
+      :- Need(
+        // A UUID string is 36 characters long,
+        // this is the string representation of the UUIDs 16 bytes.
+        |branchKeyVersionAsString| == 36,
+        Types.AwsCryptographicMaterialProvidersException(
+          message := "Branch key version string is not a UUID:" + branchKeyVersionAsString
```
texastony, line 263:
```suggestion
          message := "Branch key version string is not a UUID: " + branchKeyVersionAsString
```

texastony (approved):

texastony (approved):

seebees:
This is the encrypt time check that ensures
https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsHierarchicalKeyring.dfy#L684-L689
MUST be true on decrypt.

If the version is not a UUID that turns into 16 bytes than getting the version from the cipher text will be incorect,
and more importantly, the cipher text of the encrypted data key would be wrong as well.

texastony (approved):

texastony (approved):

PR #741, opened 2024-09-19
feat(HierarchyKeyring; CMC): Shared cache across Hierarchy Keyrings

RitvikKapila:
_Issue #, if available:_

_Description of changes:_
Duvet is expected to fail.

Adding a new type `Shared` to the union `CacheType` in order to allow a shared cache across multiple hierarchy keyrings. The customer will provide an already initialized `CryptographicMaterialsCache` as an `Shared` type, which can be shared across multiple Hierarchical Keyrings. 

As before, if any other type in CacheType is provided to the Hierarchical Keyring, it will initialize a cache of that type, to be used with only that Hierarchical Keyring. If nothing is provided here, a DefaultCache is initialized to be used with only that Hierarchical Keyring with entryCapacity = 1000.")

If the method `CreateCryptographicMaterialsCache` is called with this `Shared` `ICryptographicMaterialsCache` type as input, it MUST throw an error since the initialized `ICryptographicMaterialsCache` should already be one of the other types in the union `CacheType`, that is one out of `DefaultCache`, `NoCache`, `SingleThreadedCache`, `MultiThreadedCache`, `StormTrackingCache`.

Note that it doesn't make sense to create a shared `Shared` cache of type `NoCache` because `NoCache` type caches have entry capacity of 0.


texastony:
Ugh... No. 
We need to take the squashed commit from rc-1.7.0 and use that as the commit to PR.
Or @seebees and I will have a bad day later in our lives.

PR #751, opened 2024-09-22
chore(Mutations): respond to comments on PR #720

texastony:
Responding to @josecorella comments on #720 .

Verification failure is due to:
1. Key Store Admin's Index.dfy not proving Fresh, which has been a long standing issue with the Admin
2. In `AwsCryptographyKeyStore/test/Storage/TestGetItemsForInitializeMutation.dfy`, the test assumes `type` is in the Encryption Context. It is, but Structure.dfy does not prove that, so it must be proven else where.

PR #752, opened 2024-09-23
test(Mutations): additional testing for Mutations

texastony:
I will merge this after I give @josecorella  a chance to see if #751 addresses other concerns.

Verification failure is due to:
1. Key Store Admin's Index.dfy not proving Fresh, which has been a long standing issue with the Admin
2. In `AwsCryptographyKeyStore/test/Storage/TestGetItemsForInitializeMutation.dfy`, the test assumes `type` is in the Encryption Context. It is, but Structure.dfy does not prove that, so it must be proven else where.

A Dotnet DDB Test failed due to throttling from Nuget.


josecorella (approved):
lgtm

PR #753, opened 2024-09-23
refactor(Storage): Use TransactGet instead of BatchGet for GetInitMut

texastony:
This resolves feedback from both @seebees  and @josecorella 
on Mutations branch.

Verification failure is due to (just like in #751 ):
1. Key Store Admin's Index.dfy not proving Fresh, which has been a long standing issue with the Admin
2. In `AwsCryptographyKeyStore/test/Storage/TestGetItemsForInitializeMutation.dfy`, the test assumes `type` is in the Encryption Context. It is, but Structure.dfy does not prove that, so it must be proven else where.

A .NET MPL job failed due to .NET 48 Sig V4 Auth bug.

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
@@ -847,7 +847,7 @@ module DefaultKeyStorageInterface {
         MutationLockFromOptionalItem(ddbResponse.Responses.value[0].Item, input.Identifier);
 
       :- Need(
-        ddbResponse.Responses.value[1].Item.Some? && (|ddbResponse.Responses.value[1].Item.value| == 0),
+        ddbResponse.Responses.value[1].Item.Some? && (0 < |ddbResponse.Responses.value[1].Item.value|),
         Types.KeyStorageException(
           message:=
             "GetItemsForInitializeMutation: Could not find the ACTIVE Item. "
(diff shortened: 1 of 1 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
```diff
@@ -793,155 +806,102 @@ module DefaultKeyStorageInterface {
...
-          case _ => return Failure(
-              Types.KeyStorageException(
-                message:=
-                  "Unexpected item returned by DDB. TableName: "
-                  + ddbTableName + "\tBranch Key ID: " + input.Identifier));
-        }
-        itemIndex := 1 + itemIndex;
-      }
+      var lockItem: Option<Types.MutationLock> :-
+        MutationLockFromOptionalItem(ddbResponse.Responses.value[0].Item, input.Identifier);
```
josecorella, line 847: imo; it is easier to reason about which response item you are handling if you break them out beforehand - it is more legible and I don't have to think too hard. future you would be thankful

```
var x := ddbResponse.Responses.value[0];
var y := ddbResponse.Responses.value[1];
...
```

texastony: I can refactor this.
I even had it written this way,
but an earlier PR encouraged me to not have variables allocated that were not needed.

I took that feedback too far; 
I agree that it would be a cleaner read to identify the three items.

Question: Is the cleanest implementation to assign each type an integer?

Because then we read:
```
MutationLockFromOptionalItem(ddbResponse.Responses.value[mLockIndex].Item, input.Identifier);
```

PR #594, opened 2024-08-13
feat: Adding a storage option to the KeyStore

seebees:
The key store now allows for both a default DynamoDB table, or any custom storage system.

The important aspect about the key store
is the fact that branch keys can be versioned easily, and are cryptographically safe to use.
The actual storage medium is not important.

See: https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/changes/2024-6-17_key-store-persistance/background.md#background

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #788, opened 2024-09-28
feat(KeyStore): Key Store Admin & Storage

texastony:
### feat: Adding a storage option to the KeyStore #594 

The key store now allows for both a default DynamoDB table,
or any custom storage system.

The important aspect about the key store
is the fact that branch keys can be versioned easily,
and are cryptographically safe to use.
The actual storage medium is not important.

See: https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/changes/2024-6-17_key-store-persistance/background.md#background

### feat: Key Store Admin 

The Key Store Admin client exposes administrative
operations for the Branch Keys used by the
AWS KMS Hierarchical Keyring.

These Operations are:
- Create Key
- Version Key
- Initialize Mutation
- Apply Mutation

See the Java Examples directory for the
AWS Cryptographic Materials Provider for details:
`AwsCryptographicMaterialProviders/runtimes/java/src/examples`

### feat(KeyStore): Safely Mutating Branch Keys 

TBD


texastony:
@josecorella @seebees The Java Examples failure is because we removed the Strings from Opaque,
so we can no longer differentiate the KMS Exception.
The Runtimes are the only place we can actually assert behavior on Access Denied,
because we cannot change credentials in Dafny.

See: https://github.com/smithy-lang/smithy-dafny/issues/619

Thus, I am happy with that test failing.

Also, I would like to get this squashed into rc-1.8.0 relatively soon,
as it is a PITA to merge from MPL HEAD to rc-1.8.0 and then rebase this on rc-1.8.0.
If we do not have capacity to review,
I may just merge and look for a post-merge review.

PR #798, opened 2024-10-02
chore: reject bad storm cache configurations

ajewellamz:
_Issue #, if available:_

_Description of changes:_


AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/AwsCryptographyMaterialProvidersOperations.dfy
```diff
@@ -270,6 +291,7 @@ module AwsCryptographyMaterialProvidersOperations refines AbstractAwsCryptograph
     var cmc;
 
     if input.cache.Some? {
+      var _ := CheckCache(input.cache.value, input.ttlSeconds);
```
seebees, line 294: I think that all of these need to be `:-`

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/AwsCryptographyMaterialProvidersOperations.dfy
```diff
@@ -258,6 +259,26 @@ module AwsCryptographyMaterialProvidersOperations refines AbstractAwsCryptograph
   predicate CreateAwsKmsHierarchicalKeyringEnsuresPublicly(input: CreateAwsKmsHierarchicalKeyringInput, output: Result<IKeyring, Error>)
   {true}
 
+  function method N(n : PositiveLong) : string {
+    String.Base10Int2String(n as int)
+  }
+
+  method CheckCache(cache : CacheType, ttlSeconds: PositiveLong) returns (output : Result<bool, Error>)
```
seebees, line 266: Given that this function is returning a bool, I think that it is better modeled as an `Outcome`?

seebees (changes requested):

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/AwsCryptographyMaterialProvidersOperations.dfy
ajewellamz: Whenever I use an Outcome, I eventually get bitten by the fact that Outcomes don't compose, and so I go back to Result. Now I just always start with Result.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/AwsCryptographyMaterialProvidersOperations.dfy
ajewellamz: Whoops.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
```diff
@@ -39,11 +40,65 @@ module {:options "/functionSyntax:4" }  StormTracker {
...
+  function CheckSettings(
+    cache: Types.StormTrackingCache
+  ) : (res : Result<bool, Types.Error> )
+    ensures res.Success? ==> ConsistentSettings(cache)
+  {
+    if ConsistentSettings(cache) then
+      Success(true)
+    else
+      Failure(Types.AwsCryptographicMaterialProvidersException(message := BadCacheMsg(cache)))
+  }
```
seebees, line 99: What about this?  I also feel like we could use a `Need` that runs the error message in a thunk, what do you think?

```suggestion
  function CheckSettings(
    cache: Types.StormTrackingCache
  ) : (res : Outcome<Types.Error> )
    ensures res.Pass? ==> ConsistentSettings(cache)
  {
    if ConsistentSettings(cache) then
      Pass
    else
      Fail(Types.AwsCryptographicMaterialProvidersException(message := BadCacheMsg(cache)))
  }
```

seebees (approved):
Thanks for adding the comments. LGTM.

The change in the default worried me, but I think it is ok.

PR #846, opened 2024-10-10
fix(H-Keyring): if getCache returns Opaque Error, return Opaque

texastony:
_Issue #, if available:_

_Description of changes:_
The Cache querying logic in the H-Keyring confuses a cache miss
with an opaque error.

i.e: `getCachOutput.Failure?` is not always `EntryDoesNotExist` but could be something like `Opaque(obj? := java.nio.BufferUnderflowException`.

Thus, I suggest we refactor the methods that call the cache to differentiate b/w Opaque and `EntryDoesNotExist`.

`fix(H-Keyring): if getCache returns Opaque Error, return Opaque`


seebees (changes requested):
This is a bending change that we can not make to our customers.

seebees (approved):
Given the number of people with custom caches that exist I'm ok.

However, we need to update put, and work out updating the DB-ESDK as well

PR #858, opened 2024-10-11
chore(Python): Fetch tags as part of release

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_

Python fetches source code from within the buildspec.
For the MPL, `$COMMIT_ID` will often be a git tag.
Python will need to fetch the tags.

This isn't needed for Java/NET.
Both use `--source-version XX` in their CodeBuild config, which accepts a git tag by default.

### DIFF
--- a/AwsCryptographicMaterialProviders/codebuild/release-python/test-release.yml
+++ b/AwsCryptographicMaterialProviders/codebuild/release-python/test-release.yml
@@ -25,6 +25,7 @@ phases:
       # Switch back to the main directory
       - cd aws-cryptographic-material-providers-library
       # Assert that project version is the expected released version
+      - git fetch --tags
       - git checkout $COMMIT_ID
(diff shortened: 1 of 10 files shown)


AwsCryptographicMaterialProviders/codebuild/release-python/test-release.yml
```diff
@@ -25,6 +25,7 @@ phases:
       # Switch back to the main directory
       - cd aws-cryptographic-material-providers-library
       # Assert that project version is the expected released version
+      - git fetch --tags
```
lucasmcdonald3, line 28:
```suggestion
      - git fetch --tags 
```

AwsCryptographicMaterialProviders/codebuild/release-python/test-release.yml
```diff
@@ -25,6 +25,7 @@ phases:
...
```
lucasmcdonald3, line 28:
```suggestion
      - git fetch --tags
```

josecorella (approved):

PR #875, opened 2024-10-16
chore: Remove Dafny warnings

seebees:
Remove all Dafny warnings and set warnings to error.

By removing all the errors, `--allow-warnings` can be removed from verification. This means that any verification warning will now fail the build. This is especially relevant if `{:only}` is ever committed. See #517. This would now cause CI verification to fail.

### DIFF
--- a/SharedMakefileV2.mk
+++ b/SharedMakefileV2.mk
@@ -15,11 +15,11 @@ verify:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
 verify_single:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
 verify_service:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
 
-transpile_implementation_net: DAFNY_OPTIONS=--allow-deprecation --compile-suffix --legacy-module-names --allow-external-contracts
-transpile_test_net: DAFNY_OPTIONS=--allow-deprecation --include-test-runner --compile-suffix --legacy-module-names --allow-external-contracts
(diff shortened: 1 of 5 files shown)


SharedMakefileV2.mk
```diff
@@ -11,9 +11,9 @@ GRADLEW := ./runtimes/java/gradlew
...
 
-verify:DAFNY_OPTIONS=--allow-warnings --allow-external-contracts
-verify_single:DAFNY_OPTIONS=--allow-warnings --allow-external-contracts
-verify_service:DAFNY_OPTIONS=--allow-warnings --allow-external-contracts
+verify:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
+verify_single:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
+verify_service:DAFNY_OPTIONS=--allow-deprecation --allow-external-contracts
 
 transpile_implementation_net: DAFNY_OPTIONS=--allow-warnings --compile-suffix --legacy-module-names --allow-external-contracts
 transpile_test_net: DAFNY_OPTIONS=--allow-warnings --include-test-runner --compile-suffix --legacy-module-names --allow-external-contracts
```
josecorella, line 19:
```suggestion
transpile_implementation_net: DAFNY_OPTIONS=--allow-deprecation --compile-suffix --legacy-module-names --allow-external-contracts
transpile_test_net: DAFNY_OPTIONS=--allow-deprecation --include-test-runner --compile-suffix --legacy-module-names --allow-external-contracts
```

josecorella (approved):
small nit; lmk what you think

SharedMakefileV2.mk
seebees: I tried it! Can't take it because this will fail because we are not using `doo` files :(

PR #898, opened 2024-10-24
fix: python time externs should return integers

ajewellamz:
_Issue #, if available:_

_Description of changes:_


StandardLibrary/runtimes/python/src/smithy_dafny_standard_library/internaldafny/extern/Time.py
```diff
@@ -8,10 +8,10 @@
 # Extend generated class with our externs
 class default__(smithy_dafny_standard_library.internaldafny.generated.Time.default__):
     def CurrentRelativeTimeMilli():
-        return datetime.datetime.now(tz = pytz.UTC).timestamp() * 1000
+        return round(datetime.datetime.now(tz = pytz.UTC).timestamp() * 1000)
```
lucasmcdonald3, line 11: Should we round or truncate the milliseconds?

It looks like Java truncates but .NET rounds.

It probably doesn't matter either way

PR #948, opened 2024-11-03
chore(Java): disable verbose unit test logging

texastony:
_Issue #, if available:_

_Description of changes:_

Disable Verbose test logging to make the LocalCMCTests less flaky.
It is hard to log 300k messages to a terminal,
particularly if 10 threads are concurrently writing to the terminal.

This MAY reduce how often we encounter a LocalCMCTests failure that appears like:
```
Execution failed for task ':test'.
> Received a completed event for test with unknown id '2.25506'. Registered test ids: '[2.2, 2.3, 2.1, :test]'
```

The theory is that by removing the event lifecycle handler from the concurrency testing,
each of the 300k tests completes without triggering the handler.

The assumption is that the event lifecycle handler is a cause for concurrent testing failure.


AwsCryptographicMaterialProviders/runtimes/java/build.gradle.kts
```diff
@@ -224,18 +224,21 @@ tasks.test {
...
+//            events = mutableSetOf(org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED, org.gradle.api.tasks.testing.logging.TestLogEvent.PASSED, org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED)
+//            exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+//            showExceptions = true
+//            showCauses = true
+//            showStackTraces = true
+//            showStandardStreams = true
+//        }
+//        info.events = lifecycle.events
+//        info.exceptionFormat = lifecycle.exceptionFormat
+//    }
```
ShubhamChaturvedi7, line 241: Why not just remove this instead of commenting?

texastony: Because if a test fails,
it is very helpful to know what it is.
Which this lifecycle logging helps with.

ShubhamChaturvedi7: Then maybe use a flag or disable the params that control this? Maybe this can help: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/logging/TestExceptionFormat.html#SHORT 

I'm afraid that by commenting it out, we are creating a precedence that it's okay to comment out code for "future" use case. And have seen this in our code base. If nothing else, we can remove this and add link to the gradle doc.

ajewellamz (approved):

PR #952, opened 2024-11-04
chore(STD-Dafny): HasSubString

texastony:
_Issue #, if available:_

_Description of changes:_
HasSubString checks for a String for a Sub String.
If one is found,
Some(<index-for-start-of-sub-string>) is returned.

If not, 
None is returned.

This method was mostly written by @robin-aws 

```
chore(STD-Dafny): HasSubString
```

### DIFF
--- a/StandardLibrary/src/Sequence.dfy
+++ b/StandardLibrary/src/Sequence.dfy
@@ -7,35 +7,35 @@ module StandardLibrary.Sequence {
   // export provides SequenceEqual, StandardLibrary
   import opened UInt
 
-predicate method SequenceEqualNat<T(==)>(seq1: seq<T>, seq2: seq<T>, start1: nat, start2: nat, size: nat) : (ret : bool)
-  requires start1 + size <= |seq1|
-  requires start2 + size <= |seq2|
-  ensures ret ==> seq1[start1..start1 + size] == seq2[start2..start2 + size]
-{
-  if |seq1| > INT64_MAX_LIMIT || |seq2| > INT64_MAX_LIMIT then
-    false
-  else
-    SequenceEqual(seq1, seq2, start1 as uint64, start2 as uint64, size as uint64)
-}
-
-predicate SequenceEqual<T(==)>(seq1: seq64<T>, seq2: seq64<T>, start1: uint64, start2: uint64, size: uint64) : (ret : bool)
-  requires start1 as nat + size as nat <= |seq1| 
-  requires start2 as nat + size as nat <= |seq2|
-  ensures ret <==> seq1[start1..start1 + size] == seq2[start2..start2 + size]
-{
-  seq1[start1..start1 + size] == seq2[start2..start2 + size]
-} by method {
-  var j := start2 as uint64;
-  for i := start1 as uint64 to (start1 + size) as uint64
-    invariant j == i - start1 + start2
-    invariant forall k : uint64 | start1 <= k < i :: seq1[k] == seq2[k - start1 + start2]
+  predicate method SequenceEqualNat<T(==)>(seq1: seq<T>, seq2: seq<T>, start1: nat, start2: nat, size: nat) : (ret : bool)
+    requires start1 + size <= |seq1|
+    requires start2 + size <= |seq2|
+    ensures ret ==> seq1[start1..start1 + size] == seq2[start2..start2 + size]
   {
-    if seq1[i] != seq2[j] {
-      return false;
-    }
-    j := j + 1;
+    if |seq1| > INT64_MAX_LIMIT || |seq2| > INT64_MAX_LIMIT then
+      false
+    else
+      SequenceEqual(seq1, seq2, start1 as uint64, start2 as uint64, size as uint64)
   }
-  return true;
-}
+
+  predicate SequenceEqual<T(==)>(seq1: seq64<T>, seq2: seq64<T>, start1: uint64, start2: uint64, size: uint64) : (ret : bool)
+    requires start1 as nat + size as nat <= |seq1|
+    requires start2 as nat + size as nat <= |seq2|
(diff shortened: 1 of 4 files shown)


StandardLibrary/src/String.dfy
```diff
@@ -48,4 +50,26 @@ module StandardLibrary.String {
   {
     Int2String(n, Base10)
   }
+
+  /* Returns the index of a substraing or None, if the substring is not in the string */
+  method HasSubString(xs: string, ss: string)
+    returns (o: Wrappers.Option<nat>)
+    requires |ss| <= |xs|
```
ajewellamz, line 57: It might be more useful to skip this requirement, and just return None in this case.
Otherwise, I just have to write the `if |ss| <= |xs|`everywhere I use it.

StandardLibrary/src/String.dfy
```diff
@@ -48,4 +50,26 @@ module StandardLibrary.String {
...
+    ensures o.Some? ==>
+              o.value <= |xs| - |ss| && xs[o.value..(o.value + |ss|)] == ss
+    // TODO: ensures o.None? ==> no such index exists
+  {
+    var index: nat := 0;
+    var limit: nat := |xs| - |ss|;
+    while index <= limit
+      decreases limit - index
+    {
+      if (xs[index .. index + |ss|] == ss) {
```
ajewellamz, line 67: FYI - this is an allocation of a sequence of size |ss|

texastony: Is there a non-allocation comparison?

StandardLibrary/src/String.dfy
codyroux: indeed, the code doesn't even need to change (I think)!

StandardLibrary/src/Sequence.dfy
```diff
@@ -0,0 +1,41 @@
...
+  // export provides SequenceEqual, StandardLibrary
+  import opened UInt
+
+  predicate method SequenceEqualNat<T(==)>(seq1: seq<T>, seq2: seq<T>, start1: nat, start2: nat, size: nat) : (ret : bool)
+    requires start1 + size <= |seq1|
+    requires start2 + size <= |seq2|
+    ensures ret ==> seq1[start1..start1 + size] == seq2[start2..start2 + size]
+  {
+    if |seq1| > INT64_MAX_LIMIT || |seq2| > INT64_MAX_LIMIT then
+      false
```
texastony, line 16: Issue: False, here on line 16, signals the Sequences are not equal.

But actually, we do not know that.
All we know is the sequences are very large.

Should we change `SequenceEqualNat` to be `function method`,
such that it return a `Result<bool, Error>` and error on the sequences greater than INT64_MAX,
throwing/returning an Unsupported Operation Exception
(or some semantic equivalent of that)?
 
Or, should we move the requires from lines 22 & 23 up to here,
and say that until we/Dafny has non-INT_64 max support,
we require this?

And then the callers can decide if they want to throw an exception,
or prove that their sequences are of a reasonable size.

My preference is for moving the requires up.
I think most callers will be able to prove their sequences are smaller than 
`9,223,372,036,854,775,807`. 

For example,
I think that exceeds the number characters of any stack trace I have ever read.

The only time I think I have ever had to work with so many characters is
doing log diving across multiple services at once.

At which point,
you use dedicated to tools to execute the string search.

ajewellamz: It's unsigned, so the number is actually `18,446,744,073,709,551,616`. That's more that will fit in the RAM of any computer anywhere, and would take longer than the age of the universe to check for the substring. I guarantee that your log diving did not involve 18 exabytes of data. The Dafny team has even agreed with me, in principle, to one day generate native code with this very assumption -- that no actual sequence in a real running program will ever exceed 18,446,744,073,709,551,616 entries.

The problem with Dafny (one of the many) is that proving your sequence is smaller than 18,446,744,073,709,551,616 entries is no easier than proving that it's smaller than 42 entries.

It is perfectly valid to assume that no sequence will ever exceed 18,446,744,073,709,551,616 entries.

ajewellamz: If they can prove it, they can use `SequenceEqual` directly.
If they can't prove it, then a runtime check is needed, and better in one place than in many places.

texastony: I stand corrected.
Sorry, I had a poor understanding of the magnitude of these numbers,
and I just saw a logical/CX issue.

i.e: `false` instead of `Not Supported`. 

An alternative suggestion:
```
datatype SequenceEquality = 
    | TrueValue
    | FalseValue
    | UnSupported
 ```

This makes everything very clear.
But I am not blocking on this,
and will happily approve.

**Review by seebees — APPROVED** [review]
LGTM

File: StandardLibrary/src/Sequence.dfy [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+module StandardLibrary.Sequence {
+  // export provides SequenceEqual, StandardLibrary
+  import opened UInt
+
+  predicate method SequenceEqualNat<T(==)>(seq1: seq<T>, seq2: seq<T>, start1: nat, start2: nat, size: nat) : (ret : bool)
+    requires start1 + size <= |seq1|
+    requires start2 + size <= |seq2|
+    ensures ret ==> seq1[start1..start1 + size] == seq2[start2..start2 + size]
+  {
+    if |seq1| > INT64_MAX_LIMIT || |seq2| > UINT64_MAX_LIMIT then
```
seebees, line 15:
```suggestion
    if |seq1| > UINT64_MAX_LIMIT || |seq2| > UINT64_MAX_LIMIT then
```

seebees (changes requested):

seebees (approved):
LGTM

PR #979, opened 2024-11-12
fix(Python): CMCs release lock for unhandled runtime exceptions

lucasmcdonald3:
_Issue #, if available:_

_Description of changes:_


texastony (approved):
Nice catch.
We may want tell Dafny about this,
so they push this fix into their standard libraries Concurrent caches.

texastony (approved):

lucasmcdonald3:
> Nice catch. We may want tell Dafny about this, so they push this fix into their standard libraries Concurrent caches.

https://github.com/dafny-lang/dafny/issues/5913

PR #1011, opened 2024-11-15
feat: Storm cache supports millisecond resolution

seebees:
Under continuous failure situations having inFlightTTL and graceInterval be seconds can result in a continuous buildup of load.

The Storm Cache now supports milliseconds wait times.
When the a storm cache is empty it will also
only wait for a maximum inFlightTTL.

This means that the maximum number of serial requests
that will be waited for
before an error is now `graceInterval / inFlightTTL`.

In this situation, the cache will return/throw `InFlightTTLExceeded`.

Consider a situation where a specific branch key
will never be successfully added to the storm cache.
Under continuous load the storm cache will turn
this load from concurrent requests to serial requests.

However, if a failure is relatively fast, like 25ms,
then this serial handling of load
at a minimum of 1 request per second
will quickly cause the system to be overwhelmed.

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
@@ -195,6 +195,22 @@ module {:options "/functionSyntax:4" }  StormTracker {
         INT64_MAX_LIMIT as Types.PositiveLong
     }
 
+    lemma ExpiryTimeInSecondsCanBeConvertedToMilliseconds()
+      ensures INT64_MAX_LIMIT / 1000 == 0x20C49BA5E353F7
+    {}
+
+    predicate GracePeriod?(result: Types.GetCacheEntryOutput, now : Types.PositiveLong)
+      reads this`gracePeriod
+    {
+      // This is just showing where this constant comes from
+      // It is so that we know that we can convert
+      // the expiryTime from seconds to milliseconds
+      // This should get fixed sometime before July 29, 31252
+      ExpiryTimeInSecondsCanBeConvertedToMilliseconds();
+      && result.expiryTime < 0x20C49BA5E353F7  
+      && now < (result.expiryTime  * 1000) - gracePeriod
+    }
+
     // If entry is within `grace time` of expiration, then return EmptyFetch once per `grace interval`,
     // and return cached value otherwise
     method CheckInFlight(identifier: seq<uint8>, result: Types.GetCacheEntryOutput, now : Types.PositiveLong)
@@ -209,7 +225,7 @@ module {:options "/functionSyntax:4" }  StormTracker {
         result.expiryTime <= now / 1000
       { // expired? should be impossible
         output := CheckNewEntry(identifier, now);
-      } else if now < result.expiryTime - gracePeriod { // lots of time left
+      } else if GracePeriod?(result, now) { // lots of time left
         return Full(result);
       } else { // in grace time
         if inFlight.HasKey(identifier) {
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
@@ -127,13 +127,12 @@ module {:options "/functionSyntax:4" }  StormTracker {
     }
     var wrapped : LocalCMC.LocalCMC // the actual cache
     var inFlight: MutableMap<seq<uint8>, Types.PositiveLong> // the time at which this key became in flight
-    var gracePeriod : Types.PositiveLong // milliseconds before expiration that we start putting things in flight
-    var graceInterval : Types.PositiveLong // minimum milliseconds before putting the same key in flight again
+    var gracePeriod : Types.PositiveLong // seconds before expiration that we start putting things in flight
+    var graceInterval : Types.PositiveLong // minimum seconds before putting the same key in flight again
     var fanOut : Types.PositiveLong // maximum keys in flight at one time
-    var inFlightTTL : Types.PositiveLong // maximum milliseconds before a key is no longer in flight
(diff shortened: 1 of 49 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
```diff
@@ -122,31 +129,59 @@ module {:options "/functionSyntax:4" }  StormTracker {
...
     constructor(
-      cache: Types.StormTrackingCache
+      nameonly ghost cache: Types.StormTrackingCache,
+      nameonly entryCapacity: nat,
+      nameonly entryPruningTailSize: nat,
+      nameonly gracePeriod: Types.PositiveLong,
+      nameonly graceInterval: Types.PositiveLong,
+      nameonly fanOut: Types.PositiveLong,
+      nameonly inFlightTTL: Types.PositiveLong,
+      nameonly sleepMilli: Types.PositiveLong
```
texastony, line 148: If you can, don't do this this here.
Instead, create an Internal Datatype.
This will help us evolve this in the future.

You can then pass this around.
```
datatype InternalStormTrackerInput = 
  | InternalStormTrackerInput (
      nameonly ghost cache: Types.StormTrackingCache,
      nameonly entryCapacity: nat,
      nameonly entryPruningTailSize: nat,
      nameonly gracePeriod: Types.PositiveLong,
      nameonly graceInterval: Types.PositiveLong,
      nameonly fanOut: Types.PositiveLong,
      nameonly inFlightTTL: Types.PositiveLong,
      nameonly sleepMilli: Types.PositiveLong
 )
```

seebees: This is what `nameonly`  on the arguments does. By requiring this on the method level there is an implicit datatype that can be evolved the way an internal datatype can.

What is the advantage you see in this additional `datatype`?

seebees: I can do the mutation in the constructor. This saves updating the tests!

texastony: For me, it has helped because if I need to add or remove an argument,
or change it's type,
I only have to do it one place,
as compared to every place it is called from.

Additionally, I have been playing with adding methods to datatypes.

But your constructor solution is ideal.

texastony (approved):
LGTM, besides CI failures.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
```diff
@@ -177,23 +193,104 @@ module {:options "/functionSyntax:4" }  StormTracker {
...
+             // The + 1 comes because (now / 1000) * 1000 <= now
+             // Consider (5001 / 1000) <= 5
+             // (5001 / 1000) == 5
+             // but 5 * 1000 < 5001
+          && (result.expiryTime  * 1000) - gracePeriod <= now < (result.expiryTime + 1) * 1000
+    {
+      // This is just showing where this constant comes from
+      // It is so that we know that we can convert
+      // the expiryTime from seconds to milliseconds
+      // This should get fixed sometime before July 29, 31252
```
texastony, line 216: I love this.

texastony (approved):
LGTM

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
```diff
@@ -139,34 +143,46 @@ module {:options "/functionSyntax:4" }  StormTracker {
...
       modifies this`lastPrune, inFlight
+      requires ValidState()
+      ensures ValidState()
+      ensures
+        && old(inFlight.Size()) < fanOut as nat
+        ==>
+          && !res
+          && inFlight.content() == old(inFlight.content())
+      ensures !res ==> inFlight.Size() < fanOut as nat
+      ensures res ==> inFlight.Size() == fanOut as nat
```
ajewellamz, line 182: Not important, but `==>` could be `<==>`

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
```diff
@@ -177,23 +193,104 @@ module {:options "/functionSyntax:4" }  StormTracker {
...
+        && old(inFlight.Size()) < fanOut as nat
+        && !old(inFlight.HasKey(identifier))
+        ==> output.EmptyFetch?
+
+      // In the grace period,
+      // if fanOut has space and the identifier is inflight but has not exceed the graceInterval
+      // ==> Full
+      ensures
+        && GracePeriod?(result, now)
+        && old(inFlight.Size()) < fanOut as nat
```
ajewellamz, line 266: It seems like this shouldn't be needed. If inFlight were full, we would still return Full.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
seebees: Yup. Not too helpful as I have an `old` problem. Updated.

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/CMCs/StormTracker.dfy
seebees: No, this is an edge case :) If inFlight is full at the start, there MAY be a GC and some inflights might go away.

ajewellamz (dismissed):
LGTM

texastony (dismissed):
LGTM

texastony (approved):

PR #1012, opened 2024-11-15
fix: remove input and output traits on DynamoDB operations

ajewellamz:
_Description of changes:_

Some time ago, we switched to the smithy 2.0 DynamoDB smithy model.

Unfortunately, that comes with som baggage. Namely, all the input and output structs have the `input` and `output` traits. In smithy 2.0,  this make it illegal to use those structs in other structs, which Gazelle does for the Transforms.

It seems that removing those traits has no ill effects, and so this change allows Gazelle to continue to exist.

### DIFF
--- a/ComAmazonawsDynamodb/Model/dynamodb/model.json
+++ b/ComAmazonawsDynamodb/Model/dynamodb/model.json
@@ -798,7 +798,6 @@
         }
       },
       "traits": {
-        "smithy.api#input": {}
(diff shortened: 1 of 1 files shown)


ShubhamChaturvedi7:
Verified with a quick sanity in Go and it all seems to work there as well.

ShubhamChaturvedi7 (approved):

PR #1022, opened 2024-11-19
feat: Simplify error when decrypting a single EDK

seebees:
Decryption may attempt to decrypt multipule encrypted data keys (EDK) To return all errors we wrap them in a `CollectionOfErrors`.

However, unwrapping these errors can be complicated. To simplify, if there is only a single error,
don’t wrap it, just return the single error.


seebees:
This would change the existing error behavior of the library. This is a good idea, but it needs to bake a little more. As is this would be a breaking change.

PR #1053, opened 2024-11-25
fix: return Native Hash Algorithm instead of signature algorithm

rishav-karanjit:
_Issue #, if available:_

_Description of changes:_

The extern was using SignatureAlgorithm for OAEP encryption but we should be using a hash function instead of signature algorithm because

- `x509.SHA256WithRSA` is for signatures, not for OAEP encryption
- `rsa.EncryptOAEP` expects a hash function, not a signature algorithm

This PR also corrects the error message "dfewqfew" to a right error message.

### DIFF
--- a/AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/RSAEncryption/externs.go
+++ b/AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/RSAEncryption/externs.go
@@ -76,7 +76,7 @@ func (CompanionStruct_Default___) DecryptExtern(paddingMode AwsCryptographyPrimi
 func (CompanionStruct_Default___) EncryptExtern(paddingMode AwsCryptographyPrimitivesTypes.RSAPaddingMode, key dafny.Sequence, plainText dafny.Sequence) Wrappers.Result {
 	derPublicKey, rest := pem.Decode(dafny.ToByteArray(key))
 	if len(rest) > 0 {
-		return Wrappers.Companion_Result_.Create_Failure_(AwsCryptographyPrimitivesTypes.Companion_Error_.Create_AwsCryptographicPrimitivesError_(dafny.SeqOfChars([]dafny.Char("Invalid PEM data: Found unexpected trailing data after PEM block")...)))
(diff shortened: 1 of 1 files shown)


AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/RSAEncryption/externs.go
```diff
@@ -90,30 +91,28 @@ func (CompanionStruct_Default___) EncryptExtern(paddingMode AwsCryptographyPrimi
...
 	if err != nil {
 		return Wrappers.Companion_Result_.Create_Failure_(AwsCryptographyPrimitivesTypes.Companion_Error_.Create_AwsCryptographicPrimitivesError_(dafny.SeqOfChars([]dafny.Char(err.Error())...)))
 	}
 	return Wrappers.Companion_Result_.Create_Success_(dafny.SeqOfBytes(encryptedBytes))
 }
 
-func getNativePaddingMode(mode AwsCryptographyPrimitivesTypes.RSAPaddingMode) (x509.SignatureAlgorithm, error) {
+func getNativeHashAlgorithm(mode AwsCryptographyPrimitivesTypes.RSAPaddingMode) (crypto.Hash, error) {
 	switch mode {
-	case AwsCryptographyPrimitivesTypes.Companion_RSAPaddingMode_.Create_PKCS1_():
```
rishav-karanjit, line 106: The case for `PKCS1` is already handled in line 85 (encrypt) and line 56 (decrypt). If we get `PKCS1` here, this is an error.

rishav-karanjit:
> Detected changes to the release files or to the check-files action



> @rishav-karanjit and @rishav-karanjit, I noticed you are updating the smithy model files. Does this update need new or updated user documentation? Are you adding constraints inside list, map or union? Do you know about this issue: [smithy-lang/smithy-dafny#491](https://github.com/smithy-lang/smithy-dafny/issues/491)?

No such changes. So, I believe we can safely ignore it. Probably, something is wrong with workflow file.

AwsCryptographyPrimitives/runtimes/go/ImplementationFromDafny-go/RSAEncryption/externs.go
rishav-karanjit: Also, PKCS1 does not require hashing and returning error makes more sense in this function

PR #1089, opened 2024-12-05
test(KSA): Utilize Limit KMS Clients in Mutation D/E test

texastony:
This ensures that our implementation is correct,
as if it was not,
the incorrect KMS Client would be used,
and the request would fail.

_Issue #, if available:_

_Description of changes:_


josecorella (approved):

texastony:
> @texastony and @josecorella, I noticed you are updating the smithy model files. Does this update need new or updated user documentation? Are you adding constraints inside list, map or union? Do you know about this issue: [smithy-lang/smithy-dafny#491](https://github.com/smithy-lang/smithy-dafny/issues/491)?

We are not updating any Smithy files... I wonder why this fired... I thought I had already fixed this...

PR #1103, opened 2024-12-10
fix(MPL): remove un-used imports

texastony:
_Issue #, if available:_

_Description of changes:_

Dafny has some sensitivity with module imports that,
when this library is used in a Java environment with the DB-ESDK,
end up causing duplicated class paths.

```
[ht:validateClasspath]      for the following classes:
[ht:validateClasspath]          Sets_Compile.__default
[ht:validateClasspath]          Maps_Compile.__default
```

This occurs because the Mutation feature branch added 
a dependency on Dafny-lang's library's Maps and Sets modules,
which are already dependencies of the DB-ESDK.

Dafny's `library` flag would have avoided this,
IF and ONLY IF the DB-ESDK is compiled by Dafny 
**knowing** that the MPL uses Maps and Sets.

Because the MPL added these dependencies,
it fails class path validation with older DB-ESDKs.

We do not have a CI check for class path validation.


ajewellamz (approved):
LGTM

seebees (approved):
LGTM

PR #1071, opened 2024-12-02
Test test vector

rishav-karanjit:
_Issue #, if available:_

_Description of changes:_

_Squash/merge commit message, if applicable:_

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1126, opened 2024-12-15
chore: add new externs for file append and cpu time

ajewellamz:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
+++ b/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
@@ -83,7 +83,13 @@ pub mod DafnyLibraries {
         use std::io::Write;
         use std::path::Path;
 
-        pub fn INTERNAL_ReadBytesFromFile(
+        // Attempts to read all bytes from the file at {@code path}, and returns a tuple of the following values:
+        // isError : true iff an exception was thrown during path string conversion or when reading the file
+        // bytesRead : the sequence of bytes read from the file, or an empty sequence if {@code isError} is true
+        // errorMsg : the error message of the thrown exception if {@code isError} is true, or an empty equence otherwise
+        // We return these values individually because Result is not defined in the runtime but
+        // instead in library code. It is the responsibility of library code to construct an equivalent Result value.
+        pub(crate) fn INTERNAL_ReadBytesFromFile(
             file: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
         ) -> (
             bool,
@@ -122,15 +128,18 @@ pub mod DafnyLibraries {
             }
         }
 
+        // Get the current dirctory for use in error messages
         fn curr_dir() -> String {
             let path = std::env::current_dir();
             match path {
                 Ok(path) => format!("{}", path.display()),
-                Err(_) => "unknown".to_string(),
+                Err(_) => "Error while getting the path of current directory".to_string(),
             }
         }
 
-        pub fn INTERNAL_AppendBytesToFile(
+        // Attempts to append bytes to the file at path, creating nonexistent parent
+        // See SendBytesToFile for details
+        pub(crate) fn INTERNAL_AppendBytesToFile(
             path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
             bytes: &::dafny_runtime::Sequence<u8>,
         ) -> (
@@ -140,7 +149,9 @@ pub mod DafnyLibraries {
             SendBytesToFile(path, bytes, true)
         }
 
-        pub fn INTERNAL_WriteBytesToFile(
+        // Attempts to write bytes to the file at path, creating nonexistent parent
+        // See SendBytesToFile for details
+        pub(crate) fn INTERNAL_WriteBytesToFile(
             path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
             bytes: &::dafny_runtime::Sequence<u8>,
         ) -> (
@@ -150,10 +161,17 @@ pub mod DafnyLibraries {
             SendBytesToFile(path, bytes, false)
         }
 
-        pub fn SendBytesToFile(
+        // Attempts to write bytes to the file at path, creating nonexistent parent
+        // directories as necessary, and returns a tuple of the following values:
+        // isError : true iff an exception was thrown during path string conversion or when writing to the file
+        // errorMsg : the error message of the thrown exception if {@code isError} is true, or an empty sequence otherwise
+        // We return these values individually because {@code Result} is not defined in the runtime but
+        // instead in library code. It is the responsibility of library code to construct an equivalent Result value.
+        // if append is false, the file is truncated, otherwise we append to the existing file.
+        fn SendBytesToFile(
             path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
             bytes: &::dafny_runtime::Sequence<u8>,
-            append: bool
+            append: bool,
         ) -> (
             bool,
             ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
--- a/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
+++ b/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
@@ -130,9 +130,30 @@ pub mod DafnyLibraries {
             }
         }
 
+        pub fn INTERNAL_AppendBytesToFile(
+            path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+            bytes: &::dafny_runtime::Sequence<u8>,
+        ) -> (
+            bool,
+            ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+        ) {
+            SendBytesToFile(path, bytes, truncated)
+        }
+
         pub fn INTERNAL_WriteBytesToFile(
             path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
             bytes: &::dafny_runtime::Sequence<u8>,
+        ) -> (
+            bool,
+            ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+        ) {
+            SendBytesToFile(path, bytes, false)
+        }
+
+        pub fn SendBytesToFile(
+            path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+            bytes: &::dafny_runtime::Sequence<u8>,
+            append: bool
         ) -> (
             bool,
             ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
@@ -141,9 +162,10 @@ pub mod DafnyLibraries {
             let path = Path::new(&file_name);
 
             let maybe_file = std::fs::OpenOptions::new()
+                .append(append)
                 .write(true)
                 .create(true)
-                .truncate(true)
+                .truncate(!append)
                 .open(path);
             let mut file = match maybe_file {
                 Err(why) => {
--- a/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
+++ b/AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
@@ -137,7 +137,7 @@ pub mod DafnyLibraries {
             bool,
             ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
         ) {
-            SendBytesToFile(path, bytes, truncated)
+            SendBytesToFile(path, bytes, true)
         }
 
         pub fn INTERNAL_WriteBytesToFile(
--- a/AwsCryptographyPrimitives/runtimes/rust/Cargo.toml
+++ b/AwsCryptographyPrimitives/runtimes/rust/Cargo.toml
@@ -13,6 +13,7 @@ aws-lc-sys = "0.22.0"
 aws-smithy-runtime-api = "1.7.3"
 aws-smithy-types = "1.2.9"
 chrono = "0.4.38"
+cpu-time = "1.0.0"
 dafny_runtime = { path = "../../../smithy-dafny/TestModels/dafny-dependencies/dafny_runtime_rust"}
 dashmap = "6.1.0"
 pem = "3.0.4"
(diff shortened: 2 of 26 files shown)


StandardLibrary/src/Time.dfy
```diff
@@ -36,11 +36,19 @@ module {:extern "Time"} Time {
     // be negative.
     ensures res >= 0
 
+  // Returns the number of milliseconds of CPU used by the current process so far.
+  // Time is represented as signed over unsigned because java does not do unsigned
+  // values - net can do both so we are able to change the representation to signed.
+  method {:extern "GetCurrentCPU"} GetCurrentCPU() returns (res: int64)
```
lucasmcdonald3, line 42: I'm not a big fan of this name -- maybe `GetProcessCpuTimeMillis`?

AwsCryptographyPrimitives/runtimes/rust/Cargo.toml
```diff
@@ -13,6 +13,7 @@ aws-lc-sys = "0.22.0"
 aws-smithy-runtime-api = "1.7.3"
 aws-smithy-types = "1.2.9"
 chrono = "0.4.38"
+cpu-time = "1.0.0"
```
lucasmcdonald3, line 16: (Curious) I'm surprised you need to include this in every cargo.toml file -- is this not transitively included from StandardLibrary?

StandardLibrary/test/TestString.dfy
```diff
@@ -32,4 +34,14 @@ module TestStrings {
...
   }
-}
+
+  method {:test} TestFileIO()
+  {
+    var x :- expect WriteBytesToFile("MyFile", [1,2,3,4,5]);
+    x :- expect AppendBytesToFile("MyFile", [6,7,8,9,10]);
+    x :- expect AppendBytesToFile("MyFile", [11,12,13,14,15]);
+    var y :- expect ReadBytesFromFile("MyFile");
+    expect y == [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15];
```
lucasmcdonald3, line 44: Since this seems to be the only test of these methods, could you also add

```suggestion
    expect y == [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15];
    x :- expect WriteBytesToFile("MyFile", [1,2,3,4,5]);
    y :- expect ReadBytesFromFile("MyFile");
    expect y == [1,2,3,4,5];
```
?
Just to validate the case where a file already exists for WriteBytesToFile

lucasmcdonald3 (commented):
Python externs LGTM.

AwsCryptographyPrimitives/runtimes/rust/Cargo.toml
ajewellamz: Nope. In Dafny-Rust there is no transitive. Everything is built as one giant package, and each needs its own separate Cargo.toml.

StandardLibrary/src/Time.dfy
ajewellamz: Done

StandardLibrary/runtimes/net/STD.csproj
```diff
@@ -64,9 +64,6 @@
     <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
     <!-- Work around for dafny-lang/dafny/issues/1951; remove once resolved -->
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
-    <Compile Include="..\..\..\libraries\src\FileIO\FileIO.cs">
```
lucasmcdonald3, line 67: It looks like you got your change merged into Dafny libraries: https://github.com/dafny-lang/libraries/pull/163
Do we still want to remove our dependency on that in favor of the new `FileIO.cs`?
(I could be okay with this but just want to check)

StandardLibrary/test/TestComputeSetToOrderedSequenceUInt8Less.dfy
```diff
@@ -82,12 +84,15 @@ module TestComputeSetToOrderedSequenceUInt8Less {
   }
 
   method {:test} TestSetToOrderedSequenceManyItems() {
+    var time := Time.GetAbsoluteTime();
```
lucasmcdonald3, line 87: Is this leftover from your per-language performance test or do you want to commit this? (I'm fine either way, esp. since there are still egregious issues)

StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
```diff
@@ -1,20 +1,48 @@
...
+}
+
+func GetProcessCpuTimeMillis() int64 {
+	usage := new(syscall.Rusage)
+	syscall.Getrusage(syscall.RUSAGE_SELF, usage)
+	return (usage.Utime.Nano() + usage.Stime.Nano()) / 1000000
+
+	// var m runtime.MemStats
+	// runtime.ReadMemStats(&m)
+	// return m.TotalAlloc / 1000000
```
lucasmcdonald3, line 47:
```suggestion
```

StandardLibrary/runtimes/go/ImplementationFromDafny-go/FileIO/externs.go
```diff
@@ -25,11 +36,20 @@ func (_static CompanionStruct_Default___) INTERNAL_ReadBytesFromFile(path _dafny
 }
 
 func (_static CompanionStruct_Default___) INTERNAL_WriteBytesToFile(path _dafny.Sequence, bytes _dafny.Sequence) (isError bool, errorMsg _dafny.Sequence) {
-	p := _dafny.SequenceVerbatimString(path, false)
+	p := func() string {
```
lucasmcdonald3, line 39: I don't quite understand this change -- maybe Shubham should check this

StandardLibrary/runtimes/net/STD.csproj
ajewellamz: Yes, it seemed excessively confusing to have just one extern in just one language implemented in another repo, so now all of our externs are in one place.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/FileIO/externs.go
lucasmcdonald3: NVM -- old code was buggy and untested until https://github.com/aws/aws-cryptographic-material-providers-library/pull/1071

StandardLibrary/runtimes/net/STD.csproj
lucasmcdonald3: Agreed

StandardLibrary/test/TestComputeSetToOrderedSequenceUInt8Less.dfy
ajewellamz: Yes, I see no downside to sprinkling these into our tests here and there and my plan is to do just that.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/FileIO/externs.go
ajewellamz: Stolen from Shubham in the first place.

lucasmcdonald3 (approved):

AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
```diff
@@ -130,20 +130,42 @@ pub mod DafnyLibraries {
             }
         }
 
+        pub fn INTERNAL_AppendBytesToFile(
```
RitvikKapila, line 133: If this is an internal function, can we make it not `pub`?

RitvikKapila: Same for `INTERNAL_WriteBytesToFile`

AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
```diff
@@ -130,20 +130,42 @@ pub mod DafnyLibraries {
...
+            path: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+            bytes: &::dafny_runtime::Sequence<u8>,
+        ) -> (
+            bool,
+            ::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
+        ) {
+            SendBytesToFile(path, bytes, true)
+        }
+
         pub fn INTERNAL_WriteBytesToFile(
```
RitvikKapila, line 143: Can we copy paste the comments from Java here?
Sorry for the poor formatting. The output for Rust would be different of-course. The below code comment is just for reference and is copied from Java.

```
/// <summary>
        /// Attempts to write all given bytes to the file at the given path, creating nonexistent parent directories as necessary,
        /// and outputs the following values:
        /// <list>
        ///   <item>
        ///     <term>isError</term>
        ///     <description>
        ///       true iff an exception was thrown during path string conversion or when writing to the file
        ///     </description>
        ///   </item>
        ///   <item>
        ///     <term>errorMsg</term>
        ///     <description>
        ///       the error message of the thrown exception if <c>isError</c> is true, or an empty sequence otherwise
        ///     </description>
        ///   </item>
        /// </list>
        ///
        /// We output these values individually because Result is not defined in the runtime but instead in library code.
        /// It is the responsibility of library code to construct an equivalent Result value.
        /// </summary>
```

RitvikKapila: Same for other functions

StandardLibrary/runtimes/rust/src/dafny_libraries.rs
```diff
@@ -0,0 +1,126 @@
...
+#[allow(non_snake_case)]
+pub mod DafnyLibraries {
+    pub mod FileIO {
+        use std::fs::File;
+        use std::io::Read;
+        use std::io::Write;
+        use std::path::Path;
+
+        pub fn INTERNAL_ReadBytesFromFile(
+            file: &::dafny_runtime::Sequence<::dafny_runtime::DafnyCharUTF16>,
```
RitvikKapila, line 17: Same as the other file:
- can we make this function private?
- Can we add comments? I know other externs don't have comments so this isn't blocking, but if it is low effort and can be quickly copied from Java, we MAY want to do it.

RitvikKapila: Similarly for other functions in this file

StandardLibrary/runtimes/rust/src/dafny_libraries.rs
```diff
@@ -0,0 +1,126 @@
...
+                    dafny_runtime::Sequence::default(),
+                ),
+            }
+        }
+
+        fn curr_dir() -> String {
+            let path = std::env::current_dir();
+            match path {
+                Ok(path) => format!("{}", path.display()),
+                Err(_) => "unknown".to_string(),
```
RitvikKapila, line 59: Does this mean we can't unwrap the error from `std::env::current_dir()`? If that is the case, can we update the output message? "unknown" doesn't seem very useful. Maybe something like "Error while getting the path of current directory"?

RitvikKapila (commented):
Left a few comments. Apart from this, Rust externs look good to me.

AwsCryptographicMaterialProviders/runtimes/rust/src/dafny_libraries.rs
ajewellamz: The INTERNAL things need to be pub(crate) because they're the actual externs.
Other things can be fully private.

RitvikKapila (approved):
Thanks for making the updates. LGTM.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/FileIO/externs.go
```diff
@@ -38,3 +58,37 @@ func (_static CompanionStruct_Default___) INTERNAL_WriteBytesToFile(path _dafny.
...
+	}()
+
+	// Create directories
+	os.MkdirAll(filepath.Dir(p), os.ModePerm)
+
+	bytesArray := _dafny.ToByteArray(bytes)
+
+	f, err := os.OpenFile(p, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return true, _dafny.UnicodeSeqOfUtf8Bytes(err.Error())
```
ShubhamChaturvedi7, line 82: `_dafny.UnicodeSeqOfUtf8Bytes` might not be compatible with our implementation of dafny Strings. We expect dafny Go strings to be in `dafny.Char` while `UnicodeSeqOfUtf8Bytes` uses `CodePoint`. It's the same reason why `dafny.SequenceVerbatimString(path, false)` was failing.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/FileIO/externs.go
```diff
@@ -38,3 +58,37 @@ func (_static CompanionStruct_Default___) INTERNAL_WriteBytesToFile(path _dafny.
...
+			if notEndOfSequence {
+				s = s + string(val.(_dafny.Char))
+			} else {
+				return s
+			}
+		}
+	}()
+
+	// Create directories
+	os.MkdirAll(filepath.Dir(p), os.ModePerm)
```
ShubhamChaturvedi7, line 76: We need to handle the `error` or it might cause panic in the below lines.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
```diff
@@ -1,20 +1,44 @@
...
 func CurrentRelativeTimeMilli() int64 {
 	return time.Now().UnixMilli()
 }
+
+func (CompanionStruct_Default___) GetProcessCpuTimeMillis() int64 {
+	return GetProcessCpuTimeMillis()
+}
+
+func GetProcessCpuTimeMillis() int64 {
+	usage := new(syscall.Rusage)
```
ShubhamChaturvedi7, line 41: Not a blocker, but I prefer to be explicit about pointers. `new` will allocate a pointer to hold the type `Rusage`. Or you could just do `var rusage syscall.Rusage` and then `syscall.Getrusage(syscall.RUSAGE_SELF, &usage)` which makes obvious that we are using the address of rusage.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
```diff
@@ -1,20 +1,44 @@
...
+	syscall.Getrusage(syscall.RUSAGE_SELF, usage)
```
ShubhamChaturvedi7, line 42: The doc says this can return an error, and next line then can cause panic.

ShubhamChaturvedi7 (commented):
Left some comments on the Go externs.

PR #1133, opened 2024-12-17
fix: improve golang externs

ajewellamz:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
+++ b/StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
@@ -38,7 +38,10 @@ func (CompanionStruct_Default___) GetProcessCpuTimeMillis() int64 {
 }
 
 func GetProcessCpuTimeMillis() int64 {
-	usage := new(syscall.Rusage)
-	syscall.Getrusage(syscall.RUSAGE_SELF, usage)
+	var usage syscall.Rusage
+	err := syscall.Getrusage(syscall.RUSAGE_SELF, &usage)
+	if err != nil {
+		return 0
+	}
 	return (usage.Utime.Nano() + usage.Stime.Nano()) / 1000000
 }
--- a/StandardLibrary/test/TestString.dfy
+++ b/StandardLibrary/test/TestString.dfy
@@ -48,10 +48,11 @@ module TestStrings {
   }
 
   // ensure that FileIO error are returned properly, and not a panic! or the like
-  method {:test} TestBadFileIO()
-  {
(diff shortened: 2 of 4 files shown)


StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
```diff
@@ -38,7 +38,10 @@ func (CompanionStruct_Default___) GetProcessCpuTimeMillis() int64 {
 }
 
 func GetProcessCpuTimeMillis() int64 {
-	usage := new(syscall.Rusage)
-	syscall.Getrusage(syscall.RUSAGE_SELF, usage)
+	var usage syscall.Rusage
+	err := syscall.Getrusage(syscall.RUSAGE_SELF, &usage)
+	if err != nil {
+		return 0
```
ShubhamChaturvedi7, line 44: I assume 0 is okay, but why not return error in the method signature?

StandardLibrary/test/TestString.dfy
```diff
@@ -47,4 +47,12 @@ module TestStrings {
     expect y == [1,2,3,4,5];
   }
 
+  // ensure that FileIO error are returned properly, and not a panic! or the like
+  // This fails to fail on Windows+Dotnet, becasue \ instead of /
+  // method {:test} TestBadFileIO()
+  // {
+  //   var x := WriteBytesToFile("/../../MyFile", [1,2,3,4,5]);
```
ShubhamChaturvedi7, line 54: Why commented code? Shouldn't we remove it?

ajewellamz: 1) we really want this test, and I want to leave a reminder that we can't do this.
2) it's convenient to test locally by uncommenting, even if you can't run it in CI.

StandardLibrary/runtimes/go/ImplementationFromDafny-go/Time_/externs.go
ajewellamz: Because what else could the caller do but pretend it was zero?
None of the other languages had this issue, and so like "get current time" it seems silly to return an error.

ShubhamChaturvedi7: Yeah, Go returns error with output instead of separately? Maybe `-1` if not err?

StandardLibrary/test/TestString.dfy
ShubhamChaturvedi7: We should add a comment `Enable this test if you want to test X locally`?

PR #1195, opened 2025-01-10
feat(StdLib): Go v0.0.1 release

ShubhamChaturvedi7:
_Issue #, if available:_

_Description of changes:_
```
Initial release based on MPL 1.8.0.

Files to be reviewed:
README.md
LICENSE
CHANGELOG.md
externs (Copied over from checked-in code / Optional)
Makefile (Copied over from checked-in code / Optional)
go.mod (Copied over from checked-in code / Optional)
```
Other files are polymorphed files.


seebees (approved):
LGTM

rishav-karanjit:
Some overall things:

1. Can we have a MCM to see how this fits in? 
2. How do we know if this code runs as it ran when it was inside runtime/go? Knowing this before releasing would help MCM technician/requester?
3. What tag are we adding to do this release?

rishav-karanjit (changes requested):

rishav-karanjit:
> Mentioned in the description v0.0.1

Shouldn't it be prefixed with path from the runtime?
ref: https://go.dev/ref/mod#vcs-version

> If a module is defined in a subdirectory within the repository, that is, the [module subdirectory](https://go.dev/ref/mod#glos-module-subdirectory) portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash.

rishav-karanjit (approved):
Talked offline. The GitHub tag is prefixed with subdirectories.

Non blocking but it could be great if we could have a CI test to do go build to check if the copied code works correctly.

LGTM

PR #1204, opened 2025-01-14
fix: Adding interopt testing from JS for the Hierarchical Keyring

seebees:
These are additional test vectors from the ESDK-JS
that support the Hierarchical Keyring.
Both tests vectors for KDF CTR and the keyring itself.


seebees:
See: https://github.com/aws/aws-encryption-sdk-javascript/pull/1537

AwsCryptographyPrimitives/test/TestKDF_TestVectors.dfy
```diff
@@ -357,6 +1238,33 @@ module TestKDFK_TestVectors {
     }
   }
 
+  // It is complicated to check every byte of ever vector
+  // This takes all the okm values and digests them.
+  // This way it is easy for us to say
+  // that JS and Dafny are testing the same test vectors.
+  // I only use okm because there is a deterministic relationship
+  // between all inputs and the okm.
+  const vectorOkmDigest: seq<uint8> := [
```
texastony, line 1247: It's probably cleaner to embedded these digests into the test vectors as an additional member/field. 

Otherwise, if we change one p, we need to update this list as well.

PR #1225, opened 2025-01-16
Test orphan shape diff -- No plan to merge this PR

rishav-karanjit:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/AwsCryptographicMaterialProviders/runtimes/go/ImplementationFromDafny-go/awscryptographymaterialproviderssmithygenerated/to_dafny.go
+++ b/AwsCryptographicMaterialProviders/runtimes/go/ImplementationFromDafny-go/awscryptographymaterialproviderssmithygenerated/to_dafny.go
@@ -696,79 +696,6 @@ func MaterialProvidersConfig_ToDafny(nativeInput awscryptographymaterialprovider
 
 }
 
-func PositiveInteger_ToDafny(nativeInput int32) int32 {
-	return func() int32 {
-
-		return nativeInput
-	}()
-
-}
-
-func Materials_ToDafny(nativeInput awscryptographymaterialproviderssmithygeneratedtypes.Materials) AwsCryptographyMaterialProvidersTypes.Materials {
-	return func() AwsCryptographyMaterialProvidersTypes.Materials {
-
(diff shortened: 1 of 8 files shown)


AwsCryptographicMaterialProviders/runtimes/go/ImplementationFromDafny-go/awscryptographymaterialproviderssmithygenerated/to_dafny.go
```diff
@@ -696,6 +696,63 @@ func MaterialProvidersConfig_ToDafny(nativeInput awscryptographymaterialprovider
 
 }
 
+func Materials_ToDafny(nativeInput awscryptographymaterialproviderssmithygeneratedtypes.Materials) AwsCryptographyMaterialProvidersTypes.Materials {
```
lucasmcdonald3, line 699: This is interesting -- I wouldn't have expected this to be orphaned. Is this a duplicate conversion?

rishav-karanjit: In Go, two member shape can have same target shape and be still be different in it type conversion. One reason for this is pointer vs non pointer in the membershape. This shape might be visited as one of the memberShape but not as a target shape and their type conversion implementation could be different.

rishav-karanjit: I found a example to show the difference. Both are bytesUsed but its shape conversion is different.

```
func aws_cryptography_materialProviders_PutCacheEntryInput_bytesUsed_ToDafny(input *int32) Wrappers.Option {
	return func() Wrappers.Option {
		if input == nil {
			return Wrappers.Companion_Option_.Create_None_()
		}
		return Wrappers.Companion_Option_.Create_Some_(*input)
	}()
}
```
vs 
```
func aws_cryptography_materialProviders_UpdateUsageMetadataInput_bytesUsed_ToDafny(input int32) int32 {
	return func() int32 {

		return input
	}()
}
```

PR #1235, opened 2025-01-21
fix(KeyStoreAdmin): Exceptions for Mutations when KMS Key is Disabled

texastony:
_Issue #, if available:_

_Description of changes:_

If the KMS Key involved in a Mutation is deleted or disabled, 
along with any number of other KMS exceptions,
the Key Store Admin should attempt to classify the exception as To or From Mutation Exception,
rather than a generic Key Store Admin Exception.


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/MutationErrorRefinement.dfy
```diff
@@ -166,7 +189,23 @@ module {:options "/functionSyntax:4" } MutationErrorRefinement {
               message := "Key Management through an exception."
               + " Mutation is halted. Check access to KMS."
               + "\n" + errorContext);
-
+      }
+    }
+    if (kmsWithMsg) {
```
RitvikKapila, line 194: Both this and the last case in the `if (kmsWithMsg && knownKmsStrat) {` block (which never happens) indicate that the statement `(kmsWithMsg && knownKmsStrat) {` is always true so long as `kmsWithMessage` is true. We return some exception in each case. Are you sure that the error messages won't overlap? For instance, this line:
> var hasEncrypt? := String.HasSubString(kmsErrorMessage?.value, "Encrypt");
Wouldn't this interfere with:
```
else if (hasTerminalArn?.Some?) {
        return Types.MutationToException(
            message := "Key Management denied access to the terminal KMS Key."
            + " Mutation is halted. Check encrypt access to KMS ARN: " + terminalKmsArn + "."
            + "\n" + errorContext
          );
```
Example -- If the KMS Message has hasTerminalArn? as true, and the message says you don't have Encrypt permissions, wouldn't this be returned in
```
case "Decrypt/Encrypt" =>
```
rather than 
```
else if (hasTerminalArn?.Some?) {
```

I am also concerned that if all the if blocks in all the cases in
```
match kmsOperation {
```
are exhaustive, Line 194:
```
    if (kmsWithMsg) {
```
will never be triggered because in the previous if block `(kmsWithMsg && knownKmsStrat) {`, `knownKmsStrat` is always true.

RitvikKapila (changes requested):
Left one comment and one nit.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/KMSKeystoreOperations.dfy
```diff
@@ -102,7 +102,7 @@ module {:options "/functionSyntax:4" } KMSKeystoreOperations {
     grantTokens: KMS.GrantTokenList,
     kmsClient: KMS.IKMSClient
   )
-    returns (res: Result<KMS.GenerateDataKeyWithoutPlaintextResponse, Types.Error>)
+    returns (res: Result<KMS.GenerateDataKeyWithoutPlaintextResponse, KmsError>)
     requires kmsClient.ValidState()
```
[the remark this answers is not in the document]
seebees: No. This is a nice way to do it. `KmsError` is a sub set type. So it is only some types of `Types.Error`. This is a good mechanism to make sure that an operation only returns a specific set of errors.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/MutationErrorRefinement.dfy
texastony: I am not sure I followed all of this comment,
but I did find the logic hard to follow,
and have refactored in order to clean up the logic.

Note: We have lost some formal verification via this refactor.
In order to cleanly communicate weather we calling Decrypt or Encrypt,
I had to break the Decrypt and Encrypt calls up,
so I could know which was being handled. 

That means we have lost the formal proof that the result of Decrypt 
is passed to Encrypt.

Personally, I am fine with this loss,
as it makes the code cleaner,
and restores the purpose of `kmsOperation` to NOT be the strategy, 
but the `kmsOperation`. 

This let's me clearly label the two `if (kmsWithMsg)` blocks by what they handle.

The first block handles matching on the operation name.
The second block handles matching on the KMS ARN.

If neither block handles the error message,
then we end up in the unclassified case.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/KMSKeystoreOperations.dfy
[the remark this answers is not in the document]
texastony: As seebees says, this is NOT a breaking change.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/MutationErrorRefinement.dfy
```diff
@@ -143,32 +170,45 @@ module {:options "/functionSyntax:4" } MutationErrorRefinement {
...
+        return Types.MutationToException(
+            message := "Key Management denied access to the terminal KMS Key."
+            + " Mutation is halted. Check encrypt access to KMS ARN: " + terminalKmsArn + "."
+            + "\n" + errorContext
+          );
+      }
+    }
+    // Else we cannot match the error message by either the operation or the KMS ARN, log what we can and move on
     return Types.KeyStoreAdminException(
         message := "Key Management through an exception."
```
RitvikKapila, line 213: Nit:
```suggestion
        message := "Key Management threw an exception."
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/MutateViaDecryptEncrypt.dfy
```diff
@@ -0,0 +1,134 @@
...
+
+    var encryptResponse? := kmsClient.Encrypt(kmsEncryptRequest);
+    var encryptResponse :- encryptResponse?
+    .MapFailure(e => KMSKeystoreOperations.Types.ComAmazonawsKms(ComAmazonawsKms := e));
+
+    :- Need(
+      && encryptResponse.CiphertextBlob.Some?
+      && KMSKeystoreOperations.KMS.IsValid_CiphertextType(encryptResponse.CiphertextBlob.value),
+      KMSKeystoreOperations.Types.KeyManagementException(
+        message := "Invalid response from AWS KMS Encrypt :: KMS response's Ciphertext is invalid."
```
RitvikKapila, line 122:
```suggestion
        message := "Invalid response from AWS KMS Encrypt: KMS response's Ciphertext is invalid."
```

RitvikKapila (changes requested):
Left one comment for the `MutationErrorRefinement.dfy` file and a few nits.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/MutationErrorRefinement.dfy
```diff
@@ -143,32 +170,45 @@ module {:options "/functionSyntax:4" } MutationErrorRefinement {
                 + "\n" + errorContext
               );
           }
-        case "Decrypt/Encrypt" =>
+        case "Decrypt" =>
           var hasDecrypt? := String.HasSubString(kmsErrorMessage?.value, "Decrypt");
-          var hasEncrypt? := String.HasSubString(kmsErrorMessage?.value, "Encrypt");
```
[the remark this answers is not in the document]
texastony: I do not have the constants for KMS Strings on my right now.
I can try and find them by putting together test cases.

But I know from previous research, and from the tickets that inspired this work, that when the KMS Key is disabled or scheduled for deletion, the error message includes the KMS ARN.

However,
when the caller does not have permission to use the KMS Key with the operation (i.e: `kms:Encrypt` is not allowed), the operation name IS included in the error message, but the KMS ARN is not.

For the final case, 
the exceptions could be some opaque error such as a SigV4 error or UnknownHostException.
Which I have intentionally left as KeyStoreAdminException,
as we really have no idea what is going on.

RitvikKapila (approved):
LGTM

PR #1256, opened 2025-01-27
chore: fast fail for decrypt with keyid or alias

ajewellamz:
_Issue #, if available:_

_Description of changes:_


seebees (commented):
Yes, this is looking nice.
However, let me ask: Should we have different errors for this?
cc @texastony

ajewellamz:
> Yes, this is looking nice. However, let me ask: Should we have different errors for this? cc @texastony

I don't think this is something that someone would need to explicitly catch and act upon. This is more of a configuration error you would only see during development.

texastony (approved):

texastony:
> Yes, this is looking nice.
> 
> However, let me ask: Should we have different errors for this?
> 
> cc @texastony 

So far, we have used the services base error for illegal input. 
One day, we could break that pattern.

PR #1342, opened 2025-03-19
feat(BKS): Get* HV-2 Branch Keys

texastony:
_Issue #, if available:_

_Description of changes:_

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/ErrorMessages.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/ErrorMessages.dfy
@@ -53,4 +53,7 @@ module {:options "/functionSyntax:4" } KeyStoreErrorMessages {
     "Create key store is only supported with legacy configurations. \n" +
     "For details on how to create a DDB table manually see:\n" +
     "https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/use-hierarchical-keyring.html#hierarchical-keyring-prereqs"
+
+  const INVALID_HIERARCHY_VERSION :=
+    "Invalid hierarchy version. Expected version 1 or 2."
 }
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
@@ -8,7 +8,7 @@ include "ErrorMessages.dfy"
 include "KmsArn.dfy"
 include "../../AwsCryptographicMaterialProviders/src/CanonicalEncryptionContext.dfy"
 
-module {:options "/functionSyntax:4" } GetKeys {
+module GetKeys {
   import opened StandardLibrary
   import opened Wrappers
   import opened Seq
@@ -18,6 +18,7 @@ module {:options "/functionSyntax:4" } GetKeys {
   import DefaultKeyStorageInterface
   import KMSKeystoreOperations
   import ErrorMessages = KeyStoreErrorMessages
+  import AtomicPrimitives
 
   import Types = AwsCryptographyKeyStoreTypes
   import DDB = ComAmazonawsDynamodbTypes
@@ -170,7 +171,23 @@ module {:options "/functionSyntax:4" } GetKeys {
     var a := ToHV2EC(branchKeyItem.EncryptionContext);
     print(a);
     print("\n");
-    print(CanonicalEncryptionContext.EncryptionContextToAAD(a));
+    var b :- ParseEncryptionContext(a);
+    print(b);
+    var crypto := ProvideCryptoClient();
+    if (crypto.Failure?) {
+      var e := Types.KeyStoreException(
+        message :=
+          "yo");
+      return Failure(e);
+    }
+    var c := CanonicalEncryptionContext.EncryptionContextDigest(crypto.value, b);
+    if (c.Failure?) {
+     var e := Types.KeyStoreException(
+        message :=
+          "yo");
+      return Failure(e);
+    }
+    print(c.value);
     if (branchKeyItem.EncryptionContext[Structure.HIERARCHY_VERSION] == Structure.HIERARCHY_VERSION_1) {
       var branchKey: KMS.DecryptResponse :- KMSKeystoreOperations.DecryptKey(
         branchKeyItem,
@@ -215,7 +232,7 @@ module {:options "/functionSyntax:4" } GetKeys {
 
   method ToHV2EC(
     item: Types.EncryptionContextString
-  ) returns (output: map<string, string>)
+  ) returns (output: Types.EncryptionContextString)
   {
     var withoutReserved := set k | 
       k in item && 
@@ -236,6 +253,69 @@ module {:options "/functionSyntax:4" } GetKeys {
     return newMap;
   }
 
+  function method ParseEncryptionContext(stringEncCtx: Types.EncryptionContextString) : (res: Result<Types.EncryptionContext, Types.Error>)
+  {
+      if |stringEncCtx| == 0 then
+        Success(map[])
+      else
+        var parseResults: map<string, Result<(UTF8.ValidUTF8Bytes, UTF8.ValidUTF8Bytes), Types.Error>> :=
+          map strKey | strKey in stringEncCtx.Keys :: strKey := ParseEncryptionContextPair(strKey, stringEncCtx[strKey]);
+        if exists r | r in parseResults.Values :: r.Failure?
+        then Failure(
+              Types.KeyStoreException(message := "Encryption context contains invalid UTF8")
+            )
+        else
+          assert forall r | r in parseResults.Values :: r.Success?;
+          var utf8KeysUnique := forall k, k' | k in parseResults && k' in parseResults
+                                  :: k != k' ==> parseResults[k].value.0 != parseResults[k'].value.0;
+          if !utf8KeysUnique then Failure(Types.KeyStoreException(
+                                            message := "Encryption context keys are not unique"))  // this should never happen...
+          else Success(map r | r in parseResults.Values :: r.value.0 := r.value.1)
+  }
+
+  function method ParseEncryptionContextPair(strKey: string, strValue: string) : (res: Result<(UTF8.ValidUTF8Bytes, UTF8.ValidUTF8Bytes), Types.Error>)
+      ensures (UTF8.Encode(strKey).Success? && UTF8.Encode(strValue).Success?) <==> res.Success?
+  {
+      var key :- UTF8
+                .Encode(strKey)
+                .MapFailure(WrapStringToError);
+      var value :- UTF8
+                  .Encode(strValue)
+                  .MapFailure(WrapStringToError);
+
+      Success((key, value))
+  }
+
+  function method WrapStringToError(e: string)
+    :(ret: Types.Error)
+  {
+    Types.KeyStoreException( message := e )
+  }
+
+  method ProvideCryptoClient(
+    // Crypto?: Option<AtomicPrimitives.Types.IAwsCryptographicPrimitivesClient> := None
+    Crypto?: Option<AtomicPrimitives.AtomicPrimitivesClient> := None
+  )
+    returns (output: Result<AtomicPrimitives.AtomicPrimitivesClient, AtomicPrimitives.Types.Error>)
+    requires Crypto?.Some? ==> Crypto?.value.ValidState()
+    modifies (if Crypto?.Some? then Crypto?.value.Modifies else {})
+    ensures output.Success?
+            ==>
+              && output.value.ValidState()
+              && fresh(output.value)
+              && fresh(output.value.Modifies)
+  {
+    var Crypto: AtomicPrimitives.AtomicPrimitivesClient; //AtomicPrimitives.Types.IAwsCryptographicPrimitivesClient;
+    if (Crypto?.None?) {
+      Crypto :- AtomicPrimitives.AtomicPrimitives();
+    } else {
+      Crypto := Crypto?.value;
+    }
+    // If the customer gave us the Crypto Client, it is fresh
+    // If we create the Crypto Client, it is fresh
+    assume {:axiom} fresh(Crypto) && fresh(Crypto.Modifies);
+    return Success(Crypto);
+  }
 
   method GetBranchKeyVersion(
     input: Types.GetBranchKeyVersionInput,
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
@@ -6,6 +6,7 @@ include "Structure.dfy"
 include "KMSKeystoreOperations.dfy"
 include "ErrorMessages.dfy"
 include "KmsArn.dfy"
+include "../../AwsCryptographicMaterialProviders/src/CanonicalEncryptionContext.dfy"
 
 module {:options "/functionSyntax:4" } GetKeys {
   import opened StandardLibrary
@@ -13,6 +14,7 @@ module {:options "/functionSyntax:4" } GetKeys {
   import opened Seq
 
   import Structure
+  import CanonicalEncryptionContext
   import DefaultKeyStorageInterface
   import KMSKeystoreOperations
   import ErrorMessages = KeyStoreErrorMessages
@@ -165,7 +167,10 @@ module {:options "/functionSyntax:4" } GetKeys {
         message := ErrorMessages.INVALID_HIERARCHY_VERSION
       )
     );
-
+    var a := ToHV2EC(branchKeyItem.EncryptionContext);
+    print(a);
+    print("\n");
+    print(CanonicalEncryptionContext.EncryptionContextToAAD(a));
     if (branchKeyItem.EncryptionContext[Structure.HIERARCHY_VERSION] == Structure.HIERARCHY_VERSION_1) {
       var branchKey: KMS.DecryptResponse :- KMSKeystoreOperations.DecryptKey(
         branchKeyItem,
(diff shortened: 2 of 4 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/ErrorMessages.dfy
```diff
@@ -56,4 +56,7 @@ module {:options "/functionSyntax:4" } KeyStoreErrorMessages {
 
   const INVALID_HIERARCHY_VERSION :=
     "Invalid hierarchy version. Expected version 1 or 2."
+
+  const MD_DIGEST_SHA_NOT_MATCHED :=
+    "Decrypted md digest SHA does not match md digest sha from the table."
```
texastony, line 61: Blocking: I see what you are doing here, but no one outside of our team will have the context to understand this error message.
```suggestion
    "This Branch Key item has failed the authentication check. Either it has been tampered with or the wrong 'Logical Key Store Name' has been provided."
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
```diff
@@ -146,35 +150,111 @@ module GetKeys {
...
+      }
+      var plaintextBranchKeyWithMdDigest := branchKey.Plaintext.value;
+      :- Need(
+        |plaintextBranchKeyWithMdDigest| == Structure.AES_256_LENGTH + Structure.MD_DIGEST_LENGTH,
+        Types.KeyStoreException(
+          message := ErrorMessages.INVALID_HIERARCHY_VERSION
+        )
+      );
+      var plaintextBranchKey := plaintextBranchKeyWithMdDigest[0..Structure.AES_256_LENGTH];
+      var decryptedMdDigest := plaintextBranchKeyWithMdDigest[Structure.AES_256_LENGTH..];
```
texastony: @imabhichow looks like we are going with PDK and then MD-Digest.

texastony:
```suggestion
      var proctedMdDigest := plaintextBranchKeyWithMdDigest[Structure.AES_256_LENGTH..];
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
```diff
@@ -146,35 +150,111 @@ module GetKeys {
...
+      if (decryptedMdDigest != mdDigestSha.value) {
```
texastony:
```suggestion
      if (proctedMdDigest != mdDigestSha.value) {
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,128 @@
...
+  import Types = AwsCryptographyKeyStoreTypes
+  import KMS = ComAmazonawsKmsTypes
+  
+  import AtomicPrimitives
+  import UTF8
+  import KMSKeystoreOperations
+  import KmsArn
+  import Structure
+  
+  function method GetMdDigestFromEC(
```
texastony, line 25: Blocking: GetMdDigestFromEC implies this function returns a digest.
But it returns a map of <string,string>.

What does this method do?

It appears that it removes the "logical key store name"; 
which we could call the "storedBranchKeyContext",
so would "GetStoredBranchKeyContext" be a good name?

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,128 @@
...
+  ) : (output: Types.EncryptionContextString)
+  ensures output.Keys == item.Keys - {Structure.TABLE_FIELD}
+  ensures forall k :: k in output ==> output[k] == item[k]
+  ensures forall k :: k in output ==> k !in {Structure.TABLE_FIELD}
+  {
+    map k <- item.Keys - {Structure.TABLE_FIELD}
+      :: k := item[k]
+  }
+  
+  method GetHV2EC(
```
texastony, line 36: I get it, but for today, let's distinguish b/w the various ECs.
KMS EC :: What KMS sees.
```suggestion
  method GetHv2KmsEc(
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,128 @@
...
+    {
+        var item :| item in items;
+        items := items - { item };
+        if (|item.0| >= |Structure.ENCRYPTION_CONTEXT_PREFIX| && item.0[..|Structure.ENCRYPTION_CONTEXT_PREFIX|] == Structure.ENCRYPTION_CONTEXT_PREFIX) {
+          var newKey := item.0[|Structure.ENCRYPTION_CONTEXT_PREFIX|..];
+          newMap := newMap[newKey := item.1];
+        } else {
+          newMap := newMap[item.0 := item.1];
+        }
+    }
```
texastony, line 56: For M1, this OK... but we probably will need to refactor to a more efficient solution.
Dafny maps are not mutable... so these are new allocations (I think).
i.e: each iteration is a new allocation of the map.

There is a action we can use.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,128 @@
...
+    return withoutRestrictedField;
+  }
+
+  function method RemoveRestrictedFields(a:map<string, string>) : (output:map<string, string>)
+    ensures Structure.Hv2EncryptionContext?(output)
+  {
+    a - Structure.BRANCH_KEY_RESTRICTED_FIELD_NAMES
+  }
```
texastony, line 64: Nit: I hate bad variable names. I get it, `input` vs `a` really does not matter. 
But some part of me just does not have the flexibility to accept this.
```suggestion
  function method RemoveRestrictedFields(input:map<string, string>) : (output:map<string, string>)
    ensures Structure.Hv2EncryptionContext?(output)
  {
    input - Structure.BRANCH_KEY_RESTRICTED_FIELD_NAMES
  }
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,128 @@
...
+    then Failure(
+            Types.KeyStoreException(message := "Encryption context contains invalid UTF8")
+        )
+    else
+        assert forall r | r in parseResults.Values :: r.Success?;
+        var utf8KeysUnique := forall k, k' | k in parseResults && k' in parseResults
+                                :: k != k' ==> parseResults[k].value.0 != parseResults[k'].value.0;
+        if !utf8KeysUnique then Failure(Types.KeyStoreException(
+                                        message := "Encryption context keys are not unique"))  // this should never happen...
+        else Success(map r | r in parseResults.Values :: r.value.0 := r.value.1)
```
texastony, line 83: I think I have a helper method in BKSA for proving this.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/KMSKeystoreOperations.dfy
```diff
@@ -630,8 +641,155 @@ module {:options "/functionSyntax:4" } KMSKeystoreOperations {
...
+
+    //= aws-encryption-sdk-specification/framework/branch-key-store.md#aws-kms-branch-key-decryption
+    //= type=implication
+    //# - `CiphertextBlob` MUST be the `CiphertextBlob` attribute value on the provided EncryptedHierarchicalKey
+    && decryptRequest.CiphertextBlob == versionItem.CiphertextBlob
+
+    //= aws-encryption-sdk-specification/framework/branch-key-store.md#aws-kms-branch-key-decryption
+    //= type=implication
+    //# - `EncryptionContext` MUST be the [encryption context](#encryption-context) of the provided EncryptedHierarchicalKey
+    && decryptRequest.EncryptionContext == Some(versionItem.EncryptionContext)
```
texastony, line 781: Blocking: I suspect this is wrong.
Unless you know `versionItem.EncryptionContext` is the un-prefixed user supplied EC.

texastony (commented):
Looks good; few nits and a couple of blocks.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
imabhichow: I suggest we should use the `AwsKmsRsaKeyring` format, and I'm also following the same in Create Key -> (`Digest+PDK`) because we might use some code from `AwsKmsRsaKeyring`, I guess.

imabhichow: See if you could use this https://github.com/aws/aws-cryptographic-material-providers-library/blob/0c5cfa71c24dd75aed1651bb38caa30ebcada1bd/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsRsaKeyring.dfy#L652-L656

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -0,0 +1,181 @@
...
+    while items != {}
+        decreases |items|
+    {
+        var item :| item in items;
+        items := items - { item };
+        if (|item.0| >= |Structure.ENCRYPTION_CONTEXT_PREFIX| && item.0[..|Structure.ENCRYPTION_CONTEXT_PREFIX|] == Structure.ENCRYPTION_CONTEXT_PREFIX) {
+          var newKey := item.0[|Structure.ENCRYPTION_CONTEXT_PREFIX|..];
+          newMap := newMap[newKey := item.1];
+        } else {
+          newMap := newMap[item.0 := item.1];
```
imabhichow, line 54: It looks like this method doesn't actually remove `ENCRYPTION_CONTEXT_PREFIX`. I think we should write some tests around EC.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
```diff
@@ -146,35 +170,84 @@ module GetKeys {
...
+        branchKey.Plaintext.value
+      );
+      return Success(
+          Types.GetActiveBranchKeyOutput(
+            branchKeyMaterials := branchKeyMaterials
+          ));
+    } else if (branchKeyItemFromStorage.EncryptionContext[Structure.HIERARCHY_VERSION] == Structure.HIERARCHY_VERSION_2) {
+      // branchKeyItemFromStorage.EncryptionContext comes from storage is not the actual EC. 
+      // branchKeyItemFromStorage.EncryptionContext contains all the items in the dynamodb table and table name.
+      var hv2EC := HierarchicalVersionUtils.GetHV2EC(branchKeyItemFromStorage.EncryptionContext);
```
imabhichow, line 217: I think we can use either before calling KMS.Decrypt or Here.

```suggestion
      var hv2EC := Structure.ExtractCustomEncryptionContextAs(branchKeyItemFromStorage.EncryptionContext);
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
texastony: @imabhichow The PDK length for a BKI is always 32 bytes (256 bits), where as the RSA Keyring supports variable PDK length (really, just 3: 256, 192, & 128).

That does not mean you are wrong; but be careful about re-using to much code from the RSA keyring; the uses cases are not entirely identical.

However, we could delegate to that method and always say the PDK length is 256, assuming the branch keys never get refactored to support another size.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
rishav-karanjit: Yes it does not remove `ENCRYPTION_CONTEXT_PREFIX` if it is not in the EC (else block) but if `ENCRYPTION_CONTEXT_PREFIX` is there it will remove it (if block)

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
rishav-karanjit: The reason ExtractCustomEncryptionContextAs won't work is it filters EC that are greater than ENCRYPTION_CONTEXT_PREFIX  ([here](https://github.com/aws/aws-cryptographic-material-providers-library/blob/rishav/hv2Changes/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/Structure.dfy#L399)). So, if EC is ["Robbie":"is a dog"] then it removes this key-value which is not expected. In Get operation we only remove ENCRYPTION_CONTEXT_PREFIX if its there because there could be a customer who might be extending MPL having their own create operation. So, to untransform the EC in get operation, we remove the restricted fields and then remove the prefix only if its there.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
rishav-karanjit: But, there is definitely a bug in GetHV2EC because it is returning withoutRestrictedField  instead of newMap

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/GetKeys.dfy
```diff
@@ -146,35 +150,111 @@ module GetKeys {
...
+      var hv2BranchKey := Types.EncryptedHierarchicalKey(
+        Identifier := branchKeyItemFromStorage.Identifier,
+        Type := branchKeyItemFromStorage.Type,
+        CreateTime := branchKeyItemFromStorage.CreateTime,
+        KmsArn := branchKeyItemFromStorage.KmsArn,
+        EncryptionContext := hv2EC,
+        CiphertextBlob := branchKeyItemFromStorage.CiphertextBlob
+      );
+      var branchKey: KMS.DecryptResponse :- KMSKeystoreOperations.DecryptKeyForHV2(
```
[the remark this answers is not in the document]
rishav-karanjit: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1354/files#r2011028197

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/ErrorMessages.dfy
```diff
@@ -53,4 +53,16 @@ module {:options "/functionSyntax:4" } KeyStoreErrorMessages {
...
     "https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/use-hierarchical-keyring.html#hierarchical-keyring-prereqs"
+
+  const INVALID_HIERARCHY_VERSION :=
+    "Invalid hierarchy version. Expected version 1 or 2."
+
+  const MD_DIGEST_SHA_NOT_MATCHED :=
+    "This Branch Key item has failed the authentication check. Either it has been tampered with or the wrong 'Logical Key Store Name' has been provided."
+
+  const KMS_DECRYPT_INVALID_KEY_LENGTH_HV2 :=
+    "Invalid response from AWS KMS Decrypt: Key is not of 80 bytes. This could mean Branch Key Item in the Storage has been tampered."
```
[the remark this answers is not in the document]
rishav-karanjit: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1354/files#r2011028677

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/ErrorMessages.dfy
```diff
@@ -53,4 +53,16 @@ module {:options "/functionSyntax:4" } KeyStoreErrorMessages {
...
+
+  const INVALID_BRANCH_KEY_CONTEXT := 
+    "The branch key item is missing a required attribute. The branch key item might have been tampered to remove some attribute(s)."
```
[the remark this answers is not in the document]
rishav-karanjit: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1354/files#r2011029177

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/KMSKeystoreOperations.dfy
rishav-karanjit: Added a TODO for this predicate. Will get into it later

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
rishav-karanjit: Added a TODO for this method. I think I might need to change this method into a function so that i can use it in pre and post condition for verification.

PR #1343, opened 2025-03-19
chore: re-enable Dafny

seebees:
There is an issue with .NET and MacOS, this consolidates the building of Dafny into a single local action that we can manage.


ajewellamz (approved):
LGTM

PR #1357, opened 2025-03-25
chore(dafny): update ddb model

ajewellamz:

```
<message>
```


seebees (approved):
LGTM, but I'm not sure why Rust is not included?

ajewellamz:
Rust is not included because Rust does polymorph at build time, rather than checking in 1500 more files.

PR #1365, opened 2025-03-26
chore(java): harden LocalCMCTests.java

ajewellamz:

```
<message>
```

### DIFF
--- a/AwsCryptographicMaterialProviders/runtimes/java/src/test/LocalCMCTests.java
+++ b/AwsCryptographicMaterialProviders/runtimes/java/src/test/LocalCMCTests.java
@@ -65,7 +65,7 @@ public class LocalCMCTests {
   );
   private static final int IDS_SIZE = identifies.size();
 
-  @Test(threadPoolSize = 10, invocationCount = 300000, timeOut = 10000)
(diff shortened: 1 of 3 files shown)


AwsCryptographicMaterialProviders/runtimes/java/src/test/LocalCMCTests.java
```diff
@@ -79,37 +79,39 @@ public void TestALotOfAdding() {
...
-        .creationTime(Instant.now().getEpochSecond())
-        .expiryTime(Instant.now().getEpochSecond() + 1)
-        .materials(materials)
-        .build();
-      test.PutCacheEntry(putCacheEntryInput);
+        PutCacheEntryInput putCacheEntryInput = PutCacheEntryInput
+          .builder()
+          .identifier(cacheIdentifier)
+          .creationTime(Instant.now().getEpochSecond())
+          .expiryTime(Instant.now().getEpochSecond() + 1)
```
texastony, line 110: It's a default storm tracking cache; won't these always be in Grace Period?

Why not make the expiry time greater than the Grace period?

Or switch the cache type to multi-threaded?

Then we won't have to catch the interrupt exception, because we will never be sleeping.

texastony (approved):
Amazing!!

PR #1411, opened 2025-04-08
chore(dafny): Check terminalEC union inferredEC for unique EC 

rishav-karanjit:
### Description of changes:
This PR adds check terminalEC union inferredEC for unique EC. There is a edge case that this check will be helpful on.

Edge case:

1. I have a branch key item with unexpected EC {"Koda" : "is a dog"} which is a unique BKC
2. I mutate this item. My terminal EC is also {"Koda" : "is a dog"}.
3. While mutation, we preserve the unexpected EC `{"Koda" : "is a dog"}`.
4. Given (3.), mutating will cause collision.

Unexpected EC = attribute without key prefixed with `aws-crypto-ec`

```
chore(dafny): Check terminalEC union inferredEC for unique EC
```


rishav-karanjit:
Closing as rebased branch is here https://github.com/aws/aws-cryptographic-material-providers-library/pull/1417

PR #1421, opened 2025-04-09
chore(dafny): verify hv1 branch key item when terminal hv is 2

rishav-karanjit:

```
```

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
@@ -530,28 +530,4 @@ module {:options "/functionSyntax:4" } Mutations {
     );
     return Success(mutatedItem);
   }
-
-  // TODO-HV-2-M3: Can this method also handle mutation of hv-2 item? If so the method name must be changed.
-  method Hv1ToHv2Mutation(
-    encryptedKey: Types.AwsCryptographyKeyStoreTypes.EncryptedHierarchicalKey,
-    kmsConfiguration: Types.AwsCryptographyKeyStoreTypes.KMSConfiguration,
-    grantTokens: KMS.GrantTokenList,
-    kmsClient: KMS.IKMSClient
-  )
-    requires Structure.EncryptedHierarchicalKeyFromStorage?(encryptedKey)
-    // TODO-HV-2-M3: See if we can support hv-2 in this method itself
-    requires encryptedKey.EncryptionContext[Structure.HIERARCHY_VERSION] == Structure.HIERARCHY_VERSION_VALUE_1
-    requires KmsArn.ValidKmsArn?(encryptedKey.KmsArn)
-    requires KMSKeystoreOperations.AttemptKmsOperation?(kmsConfiguration, encryptedKey.EncryptionContext[Structure.KMS_FIELD])
-
-    requires kmsClient.ValidState()
-    modifies kmsClient.Modifies
-  {
-    var decryptRes := KMSKeystoreOperations.DecryptKeyForHv1(
-      encryptedKey,
-      kmsConfiguration,
-      grantTokens,
-      kmsClient
-    );
-  }
 }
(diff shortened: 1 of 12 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
```diff
@@ -90,6 +95,28 @@ module {:options "/functionSyntax:4" } Mutations {
     var kmsOperation: string;
     var success?: bool := false;
     var throwAwayError;
+    // TODO-HV-2-M3: Can this if condition also handle hv-2 item?
+    if (isTerminalHv2?) {
+      // TODO-HV-2-M4: Support other key manager strategy
+      :- Need(keyManagerStrategy.kmsSimple?, Types.KeyStoreAdminException(message:="only KMS Simple allow when mutating to hv-2."));
+      var decryptRes := KMSKeystoreOperations.DecryptKeyForHv1(
+        item,
+        KmsUtils.KmsSymmetricKeyArnToKMSConfiguration(Types.KmsSymmetricKeyArn.KmsKeyArn(item.KmsArn)),
```
rishav-karanjit, line 104: On re-reviewing it today. I found out that I am not considering MRK here. I will do it soon

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
```diff
@@ -23,6 +23,10 @@ module {:options "/functionSyntax:4" } Mutations {
   import MutationErrorRefinement
   import KmsUtils
 
+  datatype InitialHVVerificationState =
```
rishav-karanjit, line 26: There could be a better name then this. I will come up with something. Any suggestions?

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
```diff
@@ -23,6 +23,10 @@ module {:options "/functionSyntax:4" } Mutations {
...
+    | TerminalHV1()
+    | TerminalHV2(decryptedBranchKey: KMS.DecryptResponse)
```
texastony, line 28: Names are tough. 
If you think the name is not clear, you can always add a comment.

Though now that I think about it,
we need to call decrypt if any of the follow are true:
- strategy is decrypt/encrypt
- terminal is HV-2
- original is HV-2

I am increasingly think mutations are too damn complicated... 
if I had known that we were going to do HV-2 a year ago,
I would never have had any option by decrypt... 

For M2, let's focus on terminal HV-2;
that is what our customers need.
cc: @josecorella 

Actually... here... we can do this.
```suggestion
  datatype ActiveVerificationHolder =
    | NotDecrypt()
    | KmsDecrypt(kmsRes: KMS.DecryptResponse)
```

texastony (commented):
This looks good!
But I have realized some complications with my design,
that I am going to address via a PR.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
texastony: Good catch; we need to add some test cases to the test plan for this.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStoreAdmin/src/Mutations.dfy
```diff
@@ -90,6 +95,28 @@ module {:options "/functionSyntax:4" } Mutations {
     var kmsOperation: string;
     var success?: bool := false;
     var throwAwayError;
+    // TODO-HV-2-M3: Can this if condition also handle hv-2 item?
+    if (isTerminalHv2?) {
```
texastony, line 99: This condition is too broad;
either we semantically need `isTerminalHv2? && item.hierarchyVersion.v2?`,
or we need to create a new method that 
calls KMS Decrypt for either HV-1 or HV-2 
appropriately.

We can probably refactor `GetKeys.dfy` 
to have a generic `UnwrapBranchKeyItem` or 
`DecryptAndValidateBranchKeyItem` method
that all three of the `Get*` operations can call.

In fact, the predicate you wrote,
`ValidateKmsDecryption`,
is tantalizingly close to a generic BK Item un-wrapper,
that could collapse a significant amount of the logic in `GetKeys.dfy`
and make this logic simpler...

rishav-karanjit:
Closing in flavor of PR https://github.com/aws/aws-cryptographic-material-providers-library/pull/1442. PR https://github.com/aws/aws-cryptographic-material-providers-library/pull/1442 has same diff but to `hv-2/hv-2`

PR #1424, opened 2025-04-10
test(dafny): add tests for multiple utf8 ec entries

josecorella:
### Description of changes:

Removes TestsWithDiffUTF8Ec from tests and replaced with TestsWithMultipleUTF8Ec.
TestsWithMultipleUTF8Ec is a better test since it is multiple ec entries and triggers the sorting behavior for ESDK-JS.

```
```

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/VectorsComposition/AllRawAES.dfy
+++ b/TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/VectorsComposition/AllRawAES.dfy
@@ -171,8 +171,6 @@ module {:options "-functionSyntax:4"} AllRawAES {
     + TestsWitPsiEc
     + TestsBasicEc
     + TestControlEc
-    + TestsWithDiffUTF8Ec
     + TestsWitReplacementCharEc
-    + TestsWithOnePairOfHighCodePointUtf8ValuesInEc
     + TestsWithMultipleUTF8Ec
 }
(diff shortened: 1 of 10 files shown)


lucasmcdonald3 (commented):
I'm not sure why we're removing some of these lines.
Is this removing test coverage?

josecorella:
> Is this removing test coverage? Or are we saying we were over-testing multi-byte Unicode and can restrict the testing?

We can restrict the testing, before we were testing ec combos with only one entry like, `{"𒀀": "∰"}`. This was not testing more cases than `{"𒀀": "𒀀"}`. However; adding the new tests, `{"𒀀": "𒀀", "𓀣": "𓀣", "𓉩": "𓉩"}` trigger the sorting bug.

TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/VectorsComposition/AllRawAES.dfy
```diff
@@ -149,12 +149,28 @@ module {:options "-functionSyntax:4"} AllRawAES {
...
+          decryptDescription := keyDescription
+        )
+
   const Tests :=
     {}
     + TestsNoEc
     + TestsWitPsiEc
     + TestsBasicEc
     + TestControlEc
-    + TestsWithDiffUTF8Ec
```
lucasmcdonald3, line 158: Should we clean up the definition for this? Is it still used elsewhere?

lucasmcdonald3 (approved):

TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/VectorsComposition/AllRawAES.dfy
lucasmcdonald3: (Non-blocking)

josecorella: It is not used anywhere, we can completely remove it or leave it in for when Dafny doesnt overflow with all the tests.

PR #1428, opened 2025-04-11
wip mutate to hv2

rishav-karanjit:

```
```


rishav-karanjit (commented):
The current code considers ActiveVerificationHolder is always populated. However, this is not true most of the times and only true on active item on initialize mutation. So, there should be a code to decrypt the branch key item to verify branch key item and get the `AES-256` key while mutating to HV2. This decryption can be done everytime in while mutating to `HV-2` or only when ActiveVerificationHolder is not populated.

rishav-karanjit:
Closing for https://github.com/aws/aws-cryptographic-material-providers-library/pull/1445

PR #1427, opened 2025-04-11
fix(dafny): BKSA CreateKey formal verification

texastony:
### Squash/merge commit message, if applicable:

```
fix(dafny): BKSA Create Key HV-2

It is difficult for Dafny to keep track of the KMS Encrypt requests,
as the three are nearly identical,
with the exception of the Branch Key Context Digest (bkcd).

But a SHA-384 Digest is only "socially" injective;
the domain of SHA-384 is greater than the range.

Instead,
we can break the KMS Encrypt request into a helper
method that is much easier to reason about.

Then, 
we do not need to track the different KMS requests.
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1430, opened 2025-04-12
chore(dafny): BKS HierarchyVersionToString

texastony:

```
```


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -215,4 +214,27 @@ module {:options "/functionSyntax:4" } HierarchicalVersionUtils {
     else
       Failure("Unable to decode string")
   }
+
+  function HierarchyVersionToString(version: Types.HierarchyVersion): (output: string)
+    ensures Structure.StringIsValidHierarchyVersion?(output)
+    ensures  StringToHierarchyVersion(output) == version
```
lucasmcdonald3, line 220: nit
```suggestion
    ensures StringToHierarchyVersion(output) == version
```

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
```diff
@@ -215,4 +214,27 @@ module {:options "/functionSyntax:4" } HierarchicalVersionUtils {
...
+      case v2 => Structure.HIERARCHY_VERSION_VALUE_2
+    }
+  }
+
+  function StringToHierarchyVersion(version: string): (output: Types.HierarchyVersion)
+    requires Structure.StringIsValidHierarchyVersion?(version)
+    ensures version == Structure.HIERARCHY_VERSION_VALUE_1 ==> output.v1?
+    ensures version == Structure.HIERARCHY_VERSION_VALUE_2 ==> output.v2?
+  {
+    if version == Structure.HIERARCHY_VERSION_VALUE_1 then Types.HierarchyVersion.v1 else Types.HierarchyVersion.v2
```
lucasmcdonald3, line 233: Could/should this be safer? What happens if version is neither v1 nor v2?
I understand this passes proof due to the nature of StringIsValidHierarchyVersion, so not blocking

lucasmcdonald3 (approved):
Nits

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
texastony: Oh, so the risk is only if a user (us or an external user) manages to call the Dafny transpiled logic of this method directly, as Dafny will never allow Dafny logic to call this method without meeting the requires clause.

Personally, I am fine with this risk, as we have already accepted it everywhere we use a `requires` clause. 

Using Dafny without using `requires` is... well... it would be difficult.
I did not want to throw an error here,
as refactoring `Structure.dfy` and `MutationStateStructures.dfy` to account for an error result would be terribly expensive.

texastony:
This PR is blocked by the change requests in #1427

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/HierarchicalVersionUtils.dfy
lucasmcdonald3: I'm more thinking about the risk of us adding HV3, 4, 5... and forgetting to update this logic

texastony: The Ensures clause handle that.
Line 231 can be read out loud as:
"The version string is equal to the string '2', therefore the hierarchy-version is 2".

If we added an HV3, line 231 would fail `dafny verify`.

PR #1441, opened 2025-04-14
refactor(java): Move examples to new project to depend on ESDK

texastony:

```
refactor(java): Move examples to new project to depend on ESDK

Also, fix Java BKSA Create Example,
and add a downstream test.
```


Examples/java/src/main/java/software/amazon/cryptography/example/hierarchy/CreateKeyExample.java
```diff
@@ -59,7 +67,7 @@ public static String CreateKey(
     // this makes it easy for Crypto Tools to clean up these Example Branch Keys.
     branchKeyId =
       StringUtils.isBlank(branchKeyId)
-        ? "mpl-java-example-hv2-" + java.util.UUID.randomUUID().toString()
+        ? "mpl-java-example-" + java.util.UUID.randomUUID().toString()
```
lucasmcdonald3, line 70: I'm not a big fan of generating the UUID here. I think it's demonstrating an anti-pattern to customers. But if this unblocks something, it could be fine

Examples/java/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+var esdkVersion = "3.0.1"
+dependencies {
+    // The Library being demonstrated
+    implementation("software.amazon.cryptography:aws-cryptographic-material-providers:${mplVersion}")
+    // Dafny dependencies
+    implementation("org.dafny:DafnyRuntime:${dafnyVersion}")
+    implementation("software.amazon.smithy.dafny:conversion:0.1.1")    
+    // sdk dependencies
+    implementation(platform("software.amazon.awssdk:bom:2.26.3"))
+    // ESDK uses the MPL
```
lucasmcdonald3, line 27: Is this only used in the downstream testing? Maybe
```suggestion
    // Used in downstream testing
```

Examples/java/src/test/java/software/amazon/cryptography/example/downstream/ESDKTest.java
```diff
@@ -0,0 +1,66 @@
...
+import software.amazon.cryptography.example.hierarchy.KeyStoreProvider;
+import software.amazon.cryptography.keystore.KeyStore;
+import software.amazon.cryptography.materialproviders.IKeyring;
+import software.amazon.cryptography.materialproviders.MaterialProviders;
+import software.amazon.cryptography.materialproviders.model.CacheType;
+import software.amazon.cryptography.materialproviders.model.CreateAwsKmsHierarchicalKeyringInput;
+import software.amazon.cryptography.materialproviders.model.MaterialProvidersConfig;
+import software.amazon.cryptography.materialproviders.model.NoCache;
+
+public class ESDKTest {
```
lucasmcdonald3, line 23: This is a cool test, but
* Should this be in the examples test folder?
* I don't think it's ideal to write testing code with the ESDK inside the MPL. I thought the ideal state for downstream testing is that the MPL's CI would pull the ESDK from Maven, then run the ESDK's tests from the MPL's CI. That gets us better test coverage from the MPL and only one set of tests. But I know that is work, so this seems fine for now. But if that is the ideal state, can you add a TODO to change that?

Examples/java/src/main/java/software/amazon/cryptography/example/hierarchy/CreateKeyExample.java
texastony: I only do it if the parameter is empty;
I need the parameter to demonstrate what it does and how to use it.
I am trying to strike the middle ground of "it is clear how to create a branch key"
and "this example is a useful native End-to-End test".

Examples/java/src/test/java/software/amazon/cryptography/example/downstream/ESDKTest.java
texastony: The ESDK's tests are not in Maven... 
so it would be impossible to to the ESDK's tests without significant refactoring.

About a TODO: this branch is for HV-2. The SIM/Story I am assigned is HV-2.
I think MPL Downstream Testing is worthy SIM/Story/work,
and this is related to it (maybe as a POC),
but MPL Downstream Testing should still,
in my opinion, 
be considered as un-started/un-funded.

Hence, I am against adding a TODO... 
I like TODOs for things I am "funded" to do.

lucasmcdonald3 (approved):

PR #1448, opened 2025-04-15
chore(dafny): dont recalculate RSA key on every decrypt

ajewellamz:

```
```

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/RawRSAKeyring.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/RawRSAKeyring.dfy
@@ -113,6 +113,8 @@ module RawRSAKeyring {
       ensures this.paddingScheme == paddingScheme
       ensures this.publicKey == publicKey
       ensures ValidState() && fresh(History) && fresh(Modifies - cryptoPrimitives.Modifies)
+      ensures publicKey.None? ==> this.publicKeyMaterial.None?
+      ensures privateKey.None? ==> this.privateKeyMaterial.None?
     {
       this.keyNamespace := namespace;
       this.keyName := name;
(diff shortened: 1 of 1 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/RawRSAKeyring.dfy
```diff
@@ -162,10 +195,10 @@ module RawRSAKeyring {
...
       //# OnEncrypt MUST fail if this keyring does not have a specified [public
       //# key](#public-key).
-      ensures
-        this.publicKey.None? || |this.publicKey.Extract()| == 0
-        ==>
-          output.Failure?
+      // ensures
+      //   this.publicKey.None? || |this.publicKey.Extract()| == 0
+      //   ==>
+      //     output.Failure?
```
lucasmcdonald3, line 201: Does this verify?
```suggestion
      ensures
        publicKeyMaterial.None?
        ==>
          output.Failure?
```
I see OnDecrypt made a similar change, I'd prefer to keep the statement associated with spec

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/RawRSAKeyring.dfy
```diff
@@ -336,7 +364,7 @@ module RawRSAKeyring {
           output.Failure?
     {
       :- Need(
-        this.privateKey.Some? && |this.privateKey.Extract()| > 0,
+        privateKeyMaterial.Some?,
```
lucasmcdonald3, line 367: I see OnEncrypt uses something like
```suggestion
        privateKeyMaterial.Some? && privateKey.Some?,
```
Any reason not to align the two checks?

ajewellamz: We don't store the private key anymore.
We should change the other one to be only publicKeyMaterial.Some? because we know 
publicKey.None? ==> publicKeyMaterial.None?

AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/RawRSAKeyring.dfy
ajewellamz: ensures restored

lucasmcdonald3 (approved):
LGTM -- this doesn't get us the performance boost we were looking but this makes sense

PR #1508, opened 2025-05-09
chore(dafny): BKS CreateKey remove redundant EC checks for HV-2

texastony:
### Description of changes:
In Branch Key Creation,
until the Encryption Context 
is passed to `Structure.DecryptOnlyBranchKeyEncryptionContext`,
the Encryption Context is exactly as the customer intended it to be.

Thus, for `hierarchy-version-2`,
we do not need to do any transforms on the Encryption Context
before it is sent to KMS;
in fact, we should ensure that we do not any transforms.

```
chore(dafny): remove redundant EC checks
```

### DIFF
--- a/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/CreateKeys.dfy
+++ b/AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/CreateKeys.dfy
@@ -295,9 +295,9 @@ module {:options "/functionSyntax:4" } CreateKeys {
       ));
   }
 
-  method {:vcs_split_on_every_assert} CreateBranchAndBeaconKeysVersion2(
+  @IsolateAssertions method CreateBranchAndBeaconKeysVersion2(
     nameonly branchKeyIdentifier: string,
-    nameonly customEncryptionContext: map<string, string>,
+    nameonly encryptionContext: map<string, string>,
     nameonly timestamp: string,
     nameonly branchKeyVersion: string,
     nameonly logicalKeyStoreName: string,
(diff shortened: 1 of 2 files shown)


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/CreateKeys.dfy
```diff
@@ -295,9 +295,9 @@ module {:options "/functionSyntax:4" } CreateKeys {
       ));
   }
 
-  method {:vcs_split_on_every_assert} CreateBranchAndBeaconKeysVersion2(
+  @IsolateAssertions method CreateBranchAndBeaconKeysVersion2(
```
josecorella, line 298: use the {} syntax, unless we are using @ syntax everywhere it just looks sloppy

josecorella (changes requested):

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/CreateKeys.dfy
texastony: There are advantages to the `@` syntax over the `{}` syntax.
[From the Dafny Reference Manual 11.9](https://dafny.org/latest/DafnyRef/DafnyRef#sec-at-attributes):
> @-attributes have the same checks as regular resolved datatype values
>-    The attribute should exist
>-    Arguments should be compatible with the parameters, like for a datatype constructor call   
> 
> However, @-attributes have more checks:
>-    The attribute should be applied to a place where it can be used by Dafny
>-    Arguments should be literals

These last two bullet points indicate an IMPROVEMENT afforded by the `@` syntax.

While I appreciate style/form, I will never sacrifice improvements/function for style/form.

As an example, if the a Dafny Attribute is incorrectly spelled or used in space that is illegal, the `{}` syntax will simply be ignored.

But the `@` syntax will cause the Dafny resolver to throw an error; that is a big deal, especially to me, someone who routinely miss-spells things.

josecorella: cool, then let's compromise.
remove it from this pr and let's cut a pr that refactor {} to @

texastony: Alright, I will refactor to `{}`.

texastony: By mean I mean @imabhichow . Thank you @imabhichow !

josecorella (approved):

PR #1529, opened 2025-05-19
chore(java): demo hv-2 does not send prefixed EC in examples

rishav-karanjit:
### Description of changes:
- This is also like a test

```
```


Examples/java/src/main/java/software/amazon/cryptography/example/hierarchy/CreateKeyExample.java
```diff
@@ -92,8 +94,46 @@ public static String CreateKey(
...
+            .EncryptionContext(encryptionContextFailingCase)
+            // The Branch Key Store Admin can create HV-1 or HV-2 Branch Keys
+            .HierarchyVersion(_hierarchyVersion)
+            // But the Strategy MUST support the Hierarchy Version
+            .Strategy(strategy)
+            .build()
+        )
+        .Identifier();
+    } catch (KmsException e) {
+      // System.out.println("Exception message: " + e.getMessage());
```
rishav-karanjit, line 132: I left this as commented so that if we want to see the exception message, we can just uncomment but relooking this I think it's not worth it to leave it as commented code, anyone who wants to print will know how to print. So, I will remove this.

Examples/java/src/main/java/software/amazon/cryptography/example/hierarchy/CreateKeyExample.java
```diff
@@ -92,8 +94,45 @@ public static String CreateKey(
...
+            .HierarchyVersion(_hierarchyVersion)
+            // But the Strategy MUST support the Hierarchy Version
+            .Strategy(strategy)
+            .build()
+        )
+        .Identifier();
+    } catch (KmsException e) {
+      exceptionThrown = true;
+    }
+    assert exceptionThrown;
```
texastony, line 134: Blocking: Just move all of this to a Test.
Do not keep it in line with the examples, that confuses customers, and complicates this `CreateKey` method that we use through out the Example Testing.

texastony (changes requested):
This is really good, but let's simplify the CreateKey example and retain this integ test, by moving the integ test to the test directory.

PR #1538, opened 2025-05-21
chore: CFN for two new roles to prove prefixing/defixing behavior

texastony:

```
```

### DIFF
--- a/cfn/ESDK-Hierarchy-CI.yaml
+++ b/cfn/ESDK-Hierarchy-CI.yaml
@@ -559,3 +559,95 @@ Resources:
                 - !GetAtt GitHubCIRole.Arn
                 - "arn:aws:iam::370957321024:role/GitHub-CI-MPL-Dafny-Role-us-west-2"
             Action: "sts:AssumeRole"
+
+  KMSHV1WithPrefixedRobbie:
+    Type: "AWS::IAM::Role"
+    Properties:
+      RoleName: "KMS-HV1-With-Prefixed-Robbie"
+      Description: "Restrict to Encryption Context that contains Robbie Is a Dog."
+      Policies:
+        - PolicyName: Prefix-Robbie-Must-Exist
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "kms:Decrypt"
+                  - "kms:Encrypt"
+                  - "kms:GenerateDataKeyWithoutPlaintext"
+                  - "kms:ReEncrypt*"
+                Resource: "*"
+                Condition:
+                  StringEquals:
+                    "kms:EncryptionContext:aws-crypto-ec:Robbie": "Is a Dog."
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Principal:
+              Federated: !Sub "arn:aws:iam::${AWS::AccountId}:oidc-provider/token.actions.githubusercontent.com"
+            Action: "sts:AssumeRoleWithWebIdentity"
+            Condition:
+              StringEquals:
+                token.actions.githubusercontent.com:aud: "sts.amazonaws.com"
+              StringLike:
+                token.actions.githubusercontent.com:sub:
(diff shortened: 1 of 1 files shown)


cfn/ESDK-Hierarchy-CI.yaml
```diff
@@ -559,3 +559,101 @@ Resources:
...
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "kms:Decrypt"
+                  - "kms:Encrypt"
+                  - "kms:GenerateDataKeyWithoutPlaintext"
+                  - "kms:ReEncrypt*"
+                Resource:
+                  - !Sub "arn:aws:kms:*:${AWS::AccountId}:key/${HierarchicalGitHubKMSKeyID}"
+                  - !Sub "arn:aws:kms:*:${AWS::AccountId}:key/${HierarchicalGitHubKMSKeyIDTwo}"
```
rishav-karanjit, line 581: This key is used for HV-2 and the condition will fail.

cfn/ESDK-Hierarchy-CI.yaml
```diff
@@ -559,3 +559,101 @@ Resources:
...
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "kms:Decrypt"
+                  - "kms:Encrypt"
+                  - "kms:GenerateDataKey"
+                Resource:
+                  - !Sub "arn:aws:kms:*:${AWS::AccountId}:key/${HierarchicalGitHubKMSKeyID}"
```
rishav-karanjit, line 627: This key is used for HV-1 and the condition will fail.

rishav-karanjit:
None blocking comment: I think you want to deploy these two IAM role, assume it and test on it which is fine but I find it redundant for HV-2 as the key with logical name `KMSHV2OnlyRobbie` will be already doing that. For HV-1, there is no policy which verifies the EC is prefixed so this will be a not be redundant. 

More tests are always welcomed. So, its non blocking.

cfn/ESDK-Hierarchy-CI.yaml
rishav-karanjit: Non blocking, assuming you want a test that will fail on this condition.

cfn/ESDK-Hierarchy-CI.yaml
rishav-karanjit: Non blocking: assuming you want a test that will fail on this condition.

rishav-karanjit (dismissed):
LGTM

texastony:
> None blocking comment: I think you want to deploy these two IAM role, assume it and test on it which is fine but I find it redundant for HV-2 as the key with logical name `KMSHV2OnlyRobbie` will be already doing that. For HV-1, there is no policy which verifies the EC is prefixed so this will be a not be redundant.
> 
> More tests are always welcomed. So, its non blocking.

The goal is to Mutate from HV-1 to HV-2 and use these IAM Roles to prove the behavior of each.
Thus, I need to use a KMS Key that works for either HV-1 or HV-2.
Because the BK will start in HV-1, and then go to HV-2.

In [4a3a4fb](https://github.com/aws/aws-cryptographic-material-providers-library/commit/4a3a4fbf224e19374baa2badc2eebaf6e569479e), 
I have modified the IAM roles to except an KMS Key, 
as the behavior I am actually looking to check is irrelevant of the KMS key, 
only the KMS Encryption Context, 
and the KMS Operations used to access them.

texastony:
Successfully deployed via https://us-west-2.console.aws.amazon.com/cloudformation/home?region=us-west-2#/stacks/changesets/changes?stackId=arn%3Aaws%3Acloudformation%3Aus-west-2%3A370957321024%3Astack%2FESDK-Hierarchy-CI%2F51fc82a0-b2d3-11ed-9554-0af162fe7811&changeSetId=arn%3Aaws%3Acloudformation%3Aus-west-2%3A370957321024%3AchangeSet%2FESDK-Hiera-xsk4n45j9x-iditmrhg8sl%2F6ec43be9-dbed-4fe3-aa60-9dfa82b16778

PR #1564, opened 2025-06-16
chore(dafny): append our user agent in KMS client

rishav-karanjit:
### Description of changes:

userAgent before:
```
"userAgent": "aws-sdk-go-v2/1.32.8 ua/2.1 os/macos lang/go#1.23.9 md/GOOS#darwin md/GOARCH#arm64 api/kms#1.36.0 m/E"
```
userAgent after this PR:
```
"userAgent": "aws-sdk-go-v2/1.32.8 ua/2.1 os/macos lang/go#1.23.9 md/GOOS#darwin md/GOARCH#arm64 api/kms#1.36.0 AwsCryptographicMPL-Go-1.0.1 m/E",
```

userAgent from Java:
```
"userAgent": "aws-sdk-java/2.26.3 Mac_OS_X/15.5 OpenJDK_64-Bit_Server_VM/17.0.13+11-LTS Java/17.0.13 vendor/Amazon.com_Inc. io/sync http/Apache cfg/retry-mode/legacy cfg/auth-source#env, AwsCryptographicMPL/Java/1.0.1",
```
Note: every char not specified by ua [rule](https://github.com/aws/aws-sdk-go-v2/blob/67464865ce54f160aec17af0bae17103f7176c85/aws/middleware/user_agent.go#L370) would be converted to `-` in Go SDK

```
chore(go): append our user agent in KMS client
```

### DIFF
--- a/ComAmazonawsKms/src/Index.dfy
+++ b/ComAmazonawsKms/src/Index.dfy
@@ -3,7 +3,7 @@
 
 include "../Model/ComAmazonawsKmsTypes.dfy"
 
-module Com.Amazonaws.Kms refines AbstractComAmazonawsKmsService {
+module {:extern "software.amazon.cryptography.services.kms.internaldafny"} Com.Amazonaws.Kms refines AbstractComAmazonawsKmsService {
 
   function method DefaultKMSClientConfigType() : KMSClientConfigType {
     KMSClientConfigType


kessplas:
Should the version be `1.10.1`? Where does this version get set? Seems like it might have fallen behind.

rishav-karanjit:
> Should the version be `1.10.1`? Where does this version get set? Seems like it might have fallen behind.

I updated the version. There should be a step in MCM to update this version but until now we have never cared for it. So, I think I can safely ignore it. Also, I changed title from chore(go) to chore(dafny) because of this update. In our changelog, Dafny gets changed to "all languages".

kessplas (approved):
LGTM.

PR #1615, opened 2025-07-07
chore(CI): enable python39/310

imabhichow:

```
```

### DIFF
--- a/.github/workflows/library_codegen.yml
+++ b/.github/workflows/library_codegen.yml
@@ -28,7 +28,7 @@ jobs:
         # Note python is only used for formatting generated code
         # in this workflow.
         # The workflow installs black and docformatter and runs them on generated code.
-        python-version: [3.11]
+        python-version: [3.9]
         go-version: [1.23]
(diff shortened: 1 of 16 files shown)


imabhichow:
TestVectorsMPL [fails when run on Python 3.9.](https://github.com/aws/aws-cryptographic-material-providers-library/actions/runs/16125176583/job/45500275977?pr=1615#step:26:96)

It appears that the compiled code syntax includes a pipe operator. `Optional[bytes | bytearray]` is a syntax that was only [introduced with Python 3.10.](https://peps.python.org/pep-0604/) This could be the reason why some compiled python syntax only works with Python 3.10 and not Python 3.9.

We will not support `Python <=3.11` (or atleast `Python3.9` due to the above failure). There are also some other syntax that are not supported in `Python 3.10`.
However for the time being, until `Python 3.9` [EOL in October](https://devguide.python.org/versions/), we will not be supporting `Python3.9` &  `Python3.10` , and we may return to see if we can support `Python3.10` after that.

PR #1617, opened 2025-07-10
fix(dafny): BKS Storage GetMutation use 2x GetItem instead of TWI

texastony:

```
```


AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
```diff
@@ -585,78 +585,83 @@ module DefaultKeyStorageInterface {
...
-
-      ensures |ddbClient.History.TransactGetItems| == |old(ddbClient.History.TransactGetItems)| + 1
       ensures output.Success?
               ==>
-                && Seq.Last(ddbClient.History.TransactGetItems).output.Success?
-
+                |ddbClient.History.GetItem| == |old(ddbClient.History.GetItem)| + 2
+      ensures output.Success?
+              ==>
+                && Seq.Last(ddbClient.History.GetItem).output.Success?
```
josecorella, line 593: this will only ensure that only the last getItem succeeds - not sure that's what you want

josecorella (dismissed):
ensures could be tightened up a bit to account for both getItem calls and their constructions.
pending @seebees review

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
texastony: Yes, we could radically improve the Formal Verification of this method. I just am not sure it is worth it. I do not think formal verification is normally used as unit tests. We have many, many, integration tests that prove this method works.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
```diff
@@ -585,78 +585,83 @@ module DefaultKeyStorageInterface {
...
+            "GetMutation's GetItem for the Mutation Index returned an empty result. "
             + "Branch Key ID: " + input.Identifier
             + "\tTable Name: " + ddbTableName));
 
       /** Process sensical DDB Response */
-      var lockCanidate := ddbResponse.Responses.value[0].Item;
+      ghost var lockCanidate := commitmentRes.Item;
       var lockItem: Option<Types.MutationCommitment> :-
-        MutationCommitmentFromOptionalItem(lockCanidate, input.Identifier, ddbTableName);
+        MutationCommitmentFromOptionalItem(commitmentRes.Item, input.Identifier, ddbTableName);
```
seebees, line 659: You `Need` `commitmentRes.Item.Some? && (0 < |commitmentRes.Item.value|)` above, so you can use `StorageHelpers.MutationCommitmentFromItem` and avoid the wrapping.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
```diff
@@ -585,78 +585,83 @@ module DefaultKeyStorageInterface {
...
       assert lockItem.Some? ==> lockCanidate.Some? && Structure.MutationCommitmentAttribute?(lockCanidate.value);
 
-      var indexCanidate := ddbResponse.Responses.value[1].Item;
+      ghost var indexCanidate := indexRes.Item;
       var indexItem: Option<Types.MutationIndex> :-
-        MutationIndexFromOptionalItem(indexCanidate, input.Identifier, ddbTableName);
+        MutationIndexFromOptionalItem(indexRes.Item, input.Identifier, ddbTableName);
       assert indexItem.Some? ==> indexCanidate.Some? && Structure.MutationIndexAttribute?(indexCanidate.value);
```
seebees, line 665: Same as above, I think that this is simpler as:

```dafny
      var indexItem: Types.MutationIndex :-
        StorageHelpers.MutationIndexFromItem(indexRes.Item.value, input.Identifier, ddbTableName);
      assert Structure.MutationIndexAttribute?(indexRes.Item.value);
```

Clearly you need to wrap it back up with `Some` for your output.
Since, I don't think you want to refactor that out for this as well.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
```diff
@@ -585,78 +585,83 @@ module DefaultKeyStorageInterface {
...
-      //   && !Structure.MutationCommitmentAttribute?(Seq.Last(ddbClient.History.TransactGetItems).output.value.Responses.value[0].Item.value)
-      //   ==> output.Failure?
-      // If the index is invalid, must fail
-      // ensures
-      //   && Seq.Last(ddbClient.History.TransactGetItems).output.Success?
-      //   && Seq.Last(ddbClient.History.TransactGetItems).output.value.Responses.Some?
-      //   && |Seq.Last(ddbClient.History.TransactGetItems).output.value.Responses.value| == 2
-      //   && Seq.Last(ddbClient.History.TransactGetItems).output.value.Responses.value[1].Item.Some?
-      //   && !Structure.MutationIndexAttribute?(Seq.Last(ddbClient.History.TransactGetItems).output.value.Responses.value[1].Item.value)
-      //   ==> output.Failure?
```
seebees, line 613: I think that what you want here now this?

```dafny
      ensures
          && |ddbClient.History.GetItem| == |old(ddbClient.History.GetItem)| + 2
          && Seq.Last(ddbClient.History.GetItem).output.Success?
          && Seq.Last(ddbClient.History.GetItem).output.value.Item.Some?
          && 0 < |Seq.Last(ddbClient.History.GetItem).output.value.Item.value|
          && Seq.Last(Seq.DropLast(ddbClient.History.GetItem)).output.Success?
          && Seq.Last(Seq.DropLast(ddbClient.History.GetItem)).output.value.Item.Some?
          && 0 < |Seq.Last(Seq.DropLast(ddbClient.History.GetItem)).output.value.Item.value|
          && (
            || !Structure.MutationCommitmentAttribute?(Seq.Last(Seq.DropLast(ddbClient.History.GetItem)).output.value.Item.value)
            || !Structure.MutationIndexAttribute?(Seq.Last(ddbClient.History.GetItem).output.value.Item.value)
          )
        ==> output.Failure?
```

seebees (changes requested):
I'm not worried about the verification, though I think I have that right.
But I think doing less wrapping is simple and worth it :)

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
texastony: I think we learned the hard way that prooving failure is too difficult.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
texastony: FYI: this method can be called and succeed with an empty result (no Mutation in-flight, for Describe Mutation).
I do not think we need to un-wrap these...

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
texastony: This method is used by Describe Mutation, which can return Nones if there is no Mutation in-flight.
That is why it is written this way; a none-result is valid.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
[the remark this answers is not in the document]
texastony: Eh. I will take it, but the output can be a success if the items do not exist.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
texastony: Also, Describe Mutation and Initialize Mutation will work if only the commitment exists, and assume the index was deleted to restart the mutation from the head of the table.

AwsCryptographicMaterialProviders/dafny/AwsCryptographyKeyStore/src/DefaultKeyStorageInterface.dfy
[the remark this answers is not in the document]
seebees: It that is true, then it won't prove.
But as I comment below, I think that it will.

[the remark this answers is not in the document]
texastony: I see what happened.
TGI ALWAYS returns some results, but the results can be none.
GetItem is different... looking into it.

seebees (approved):
LGTM

PR #1628, opened 2025-07-17
fix(python): add TODOs for tracking work moving forward

ShreyesS:
### Changes

Added changes to keep track of work to complete moving forward. Search up `TODO-Fuzztesting` in fuzz_generator.py to locate WIP.

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/runtimes/python/fuzz_generator.py
+++ b/TestVectorsAwsCryptographicMaterialProviders/runtimes/python/fuzz_generator.py
@@ -7,19 +7,6 @@ Fuzz Testing Generator for AWS Cryptographic Material Providers Library
 This script generates fuzzed test vectors with a focused approach:
 1. Fuzz encryption context fields and key-related fields (strings; since there a variety of Unicode characters that can be tested)
 2. Randomly choosing and testing the "representative values" -> https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/test-vectors/test-vector-enumeration.md#selecting-a-representative-input-value
-
-#TODO-Fuzztesting:
-- "negative-encrypt-keyring" fuzzing functionality: currently, we implement tests with missing required keys (for KMS keyrings) or invalid key material (raw keryings)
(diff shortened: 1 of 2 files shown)


TestVectorsAwsCryptographicMaterialProviders/runtimes/python/fuzz_generator.py
```diff
@@ -52,7 +39,10 @@
 KMS_KEYS = ["us-west-2-mrk", "us-east-1-mrk", "us-west-2-decryptable", "us-west-2-encrypt-only"]
 MRK_KEYS = ["us-west-2-mrk", "us-east-1-mrk"]
 
+#TODO-Fuzztesting: #include 𝟁-nonascii-𐀂-aes-256-𝟁-with-� and rsa for raw keyrings
```
rishav-karanjit, line 42: Isn't there any other raw keyrings missing except 𝟁-nonascii-𐀂-aes-256-𝟁-with-� and rsa?

TestVectorsAwsCryptographicMaterialProviders/runtimes/python/keys.json
rishav-karanjit, line 1: This file is likely a redundant one. If you like you can check with Lucas and remove it from main.

ShreyesS: I wanted to add it because I removed it from the last merge.

TestVectorsAwsCryptographicMaterialProviders/runtimes/python/fuzz_generator.py
rishav-karanjit: Nit: You can also say "TODO: include all raw key types from keys.json".

TestVectorsAwsCryptographicMaterialProviders/runtimes/python/keys.json
rishav-karanjit: So, does this mean that this file is not redundant?

rishav-karanjit: Note: This is non blocking

PR #1634, opened 2025-07-23
chore(python): create custom Makefile commands for fuzzing test vectors

ShreyesS:
## Changes 
Created 2 Makefile commands to support test vector fuzzing, and will be used in the CI.

### DIFF
--- a/TestVectorsAwsCryptographicMaterialProviders/Makefile
+++ b/TestVectorsAwsCryptographicMaterialProviders/Makefile
@@ -254,3 +254,31 @@ test_decrypt_encrypt_vectors_rust:
 
 test_decrypt_encrypt_vectors_go:
 	go -C runtimes/go/ImplementationFromDafny-go run ImplementationFromDafny.go decrypt --manifest-path=..
+
+# Fuzz Testing Commands
+
+# Generate fuzzed test vectors in the Python runtime directory (50 vectors)
+test_generate_fuzz_vectors:
+    cp dafny/TestVectorsAwsCryptographicMaterialProviders/test/keys.json runtimes/python/
+    cd runtimes/python && python3 fuzz_generator.py --num-vectors 50
+    @echo "fuzzed test vectors generated in runtimes/python/"
+
+# Run fuzz interoperability test from one language to another
+# Usage: make test_fuzz_interop ENCRYPT_LANG=go DECRYPT_LANG=python
+test_fuzz_interop:
+    @echo "Testing interoperability from $(ENCRYPT_LANG) to $(DECRYPT_LANG)"
+    $(MAKE) test_generate_fuzz_vectors
+    @if [ "$(ENCRYPT_LANG)" != "python" ]; then \
+        echo "Copying test vectors to $(ENCRYPT_LANG) runtime"; \
+        cp runtimes/python/manifest.json runtimes/$(ENCRYPT_LANG)/; \
+        cp runtimes/python/keys.json runtimes/$(ENCRYPT_LANG)/; \
+    fi
+    @echo "Encrypting with $(ENCRYPT_LANG) runtime"
+    $(MAKE) test_encrypt_vectors_$(ENCRYPT_LANG)
+    @if [ "$(ENCRYPT_LANG)" != "$(DECRYPT_LANG)" ]; then \
+        echo "Copying encrypted manifest and keys to $(DECRYPT_LANG) runtime"; \
+        cp runtimes/$(ENCRYPT_LANG)/manifest.json runtimes/$(DECRYPT_LANG)/; \
+        cp runtimes/$(ENCRYPT_LANG)/keys.json runtimes/$(DECRYPT_LANG)/; \
+    fi
+    @echo "Decrypting with $(DECRYPT_LANG) runtime"
+    $(MAKE) test_decrypt_encrypt_vectors_$(DECRYPT_LANG)


TestVectorsAwsCryptographicMaterialProviders/Makefile
```diff
@@ -254,3 +254,31 @@ test_decrypt_encrypt_vectors_rust:
...
+
+# Generate fuzzed test vectors in the Python runtime directory (50 vectors)
+test_generate_fuzz_vectors:
+    cp dafny/TestVectorsAwsCryptographicMaterialProviders/test/keys.json runtimes/python/
+    cd runtimes/python && python3 fuzz_generator.py --num-vectors 50
+    @echo "fuzzed test vectors generated in runtimes/python/"
+
+# Run fuzz interoperability test from one language to another
+# Usage: make test_fuzz_interop ENCRYPT_LANG=go DECRYPT_LANG=python
+test_fuzz_interop:
```
rishav-karanjit, line 268: Interop test does not need this why do we need this for interoperability for fuzzed test?

TestVectorsAwsCryptographicMaterialProviders/Makefile
```diff
@@ -254,3 +254,31 @@ test_decrypt_encrypt_vectors_rust:
...
 test_decrypt_encrypt_vectors_go:
 	go -C runtimes/go/ImplementationFromDafny-go run ImplementationFromDafny.go decrypt --manifest-path=..
+
+# Fuzz Testing Commands
+
+# Generate fuzzed test vectors in the Python runtime directory (50 vectors)
+test_generate_fuzz_vectors:
+    cp dafny/TestVectorsAwsCryptographicMaterialProviders/test/keys.json runtimes/python/
+    cd runtimes/python && python3 fuzz_generator.py --num-vectors 50
+    @echo "fuzzed test vectors generated in runtimes/python/"
```
rishav-karanjit, line 264: This echo will only add noise. When this command succeeds its clear it has generated

TestVectorsAwsCryptographicMaterialProviders/Makefile
```diff
@@ -254,3 +254,31 @@ test_decrypt_encrypt_vectors_rust:
 
 test_decrypt_encrypt_vectors_go:
 	go -C runtimes/go/ImplementationFromDafny-go run ImplementationFromDafny.go decrypt --manifest-path=..
+
+# Fuzz Testing Commands
+
+# Generate fuzzed test vectors in the Python runtime directory (50 vectors)
+test_generate_fuzz_vectors:
+    cp dafny/TestVectorsAwsCryptographicMaterialProviders/test/keys.json runtimes/python/
+    cd runtimes/python && python3 fuzz_generator.py --num-vectors 50
```
rishav-karanjit, line 263: 50 is too less. Given the current interop generates 1000+.

TestVectorsAwsCryptographicMaterialProviders/Makefile
```diff
@@ -254,3 +254,31 @@ test_decrypt_encrypt_vectors_rust:
 
 test_decrypt_encrypt_vectors_go:
 	go -C runtimes/go/ImplementationFromDafny-go run ImplementationFromDafny.go decrypt --manifest-path=..
+
+# Fuzz Testing Commands
+
+# Generate fuzzed test vectors in the Python runtime directory (50 vectors)
+test_generate_fuzz_vectors:
```
rishav-karanjit, line 261: The name of the command does not differentiate if its python or other runtimes.
```suggestion
test_generate_fuzz_vectors_python:
```

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: This is blocking because we will be have to maintain two different process where we can have just one.

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: This is blocking too. We either need to remove this if it does not add any value. If it adds value then add it to all the existing make file commands.

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: We can put `fuzz_generator.py` in a common directory where we put dafny code because we use this for every runtime.

ShreyesS: I chose 50 as a starting point. With the final CI, I'll ensure we can run tens of thousands of tests.

Also, I'll make a TODO for the common directory. I would like to discuss the logic of the workflow with you.

TestVectorsAwsCryptographicMaterialProviders/Makefile
ShreyesS: The name might be misleading, but essentially I need the setup and runtime implementations that are done in interop to run cross-runtime testing.

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: Yeah sure. If you have a TODO its fine but don't let all the TODO drown you. You goal is to have this running in main not in any other branch

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: > I chose 50 as a starting point. With the final CI, I'll ensure we can run tens of thousands of tests.

Have a TODO for this. I want to set the variable such that it runs for atleast 6 hours

rishav-karanjit: > Also, I'll make a TODO for the common directory. I would like to discuss the logic of the workflow with you.

Sure but don't create TODO for every thing that it gets very difficult for you to merge to main in the end. Currently, all the stuffs in `runtimes/<language>` is used by that language itself like Go does not need anything in `runtimes/python`.

TestVectorsAwsCryptographicMaterialProviders/Makefile
rishav-karanjit: With this command, its hard to re-use the existing CI workflow. I am fine to check in this code if you want to get your hand dirty and look around but its unlikely that you will re-use existing interop CI and use this command.

PR #1653, opened 2025-07-30
chore(go): update Go release script for ESDK and DB-ESDK

rishav-karanjit:
### Description of changes:
- The plan is DB-ESDK, ESDK and MPL will use the script in MPL so that we don't have to maintain in 3 places. I need to remove the script that's already checked in DB-ESDK and make DB-ESDK use the script in MPL.

- Tested by triggering the workflow: https://github.com/aws/aws-cryptographic-material-providers-library/actions/runs/16627856159

```
chore(go): update Go release script for ESDK and DB-ESDK
```

### DIFF
--- a/scripts/go-release-automation.sh
+++ b/scripts/go-release-automation.sh
@@ -69,7 +69,7 @@ run_release_script() {
   run_go_tools
 
   # Replacement directives are removed to get package from go pkg instead of local copy
-  # We need to copy `go.mod` otherwise without it this script will not work in the first release because all the go tools used in this script works only when there is a `go.mod` in the directory.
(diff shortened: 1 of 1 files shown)


ajewellamz (approved):

scripts/go-release-automation.sh
```diff
@@ -64,17 +66,11 @@ run_release_script() {
...
 
-  echo "Running go get -u..."
-  go get -u
-
-  echo "Running go mod tidy..."
-  go mod tidy
-
-  echo "Running go build to check for errors..."
-  go build --gcflags="-e" ./...
+  # Replacement directives are removed to get package from go pkg instead of local copy
```
rishav-karanjit, line 71:
```suggestion
  # Replacement directives are removed to get package from go pkg instead of local copy
  # We need to copy `go.mod` otherwise without it this script will not work in the first release because all the go tools used in this script works only when there is a `go.mod` in the directory.
```

PR #1677, opened 2025-08-07
chore(python): fuzz pr ci and daily ci

ShreyesS:
## Changes
- PR CI and Daily CI

## WIP
- self review
- Quip Doc of Fuzz-testing (will include TODOs there)
- removing TODOs from Script

```
```

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -67,4 +67,10 @@ jobs:
     uses: ./.github/workflows/library_interop_tests.yml
     with:
       dafny: ${{needs.getVersion.outputs.version}}
+  daily-fuzz-interop-test:
+    needs: getVersion
+    if: github.event_name != 'schedule' || github.repository_owner == 'aws'
+    uses: ./.github/workflows/library_interop_tests.yml
+    with:
+      dafny: ${{needs.getVersion.outputs.version}}
       fuzz-testing: true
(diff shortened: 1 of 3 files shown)


.github/workflows/daily_ci.yml
```diff
@@ -67,3 +67,4 @@ jobs:
     uses: ./.github/workflows/library_interop_tests.yml
     with:
       dafny: ${{needs.getVersion.outputs.version}}
+      fuzz-testing: true
```
rishav-karanjit, line 70: We don't want to remove current test vector but want to add fuzz testing. This means there will be fuzz testing only

TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/test/fuzz_generator.py
```diff
@@ -325,7 +321,6 @@ def generate_fuzz_test_vectors(num_vectors) -> Tuple[Dict[str, Any], Dict[str, A
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=NonInteractiveExampleWarning)
         for i in range(num_vectors):
-            # TODO-Fuzztesting: remove .example() usage. Context: we're using Hypothesis as a data generator, not for testing properties.
             # Hypothesis is designed for property-based testing, so when using .example() it informs us that we should be using @given to actually test properties, not just generate examples.
```
rishav-karanjit, line 324: Let's add a TODO because I still want anyone looking into this fuzz generator to know this. TODOs are generally highlighted.
```suggestion
            # TODO: remove .example() usage.
            # Hypothesis is designed for property-based testing, so when using .example() it informs us that we should be using @given to actually test properties, not just generate examples.
```

PR #1686, opened 2025-08-14
chore(python): add user agent suffix to kms requests

imabhichow:
### Description of changes:

Before Change:
      
  `"userAgent": "Boto3/1.40.10 md/Botocore#1.40.10 ua/2.1 os/macos#24.5.0 md/arch#arm64 lang/python#3.12.11 md/pyimpl#CPython m/Z,b,D cfg/retry-mode#legacy Botocore/1.40.10"`

After Change

 `"userAgent": "Boto3/1.40.10 md/Botocore#1.40.10 ua/2.1 os/macos#24.5.0 md/arch#arm64 lang/python#3.12.11 md/pyimpl#CPython m/b,D,Z cfg/retry-mode#legacy Botocore/1.40.10 AwsCryptographicMPL/Python/1.11.0"`

```
chore(python): add user agent suffix to kms requests
```

### DIFF
--- a/.releaserc.cjs
+++ b/.releaserc.cjs
@@ -372,6 +372,15 @@ module.exports = {
             ),
             countMatches: true,
           },
+
+          // Update the version in ComAmazonawsKms/src/Index.dfy DafnyUserAgentSuffix function
+          {
+            files: ["ComAmazonawsKms/src/Index.dfy"],
+            from: 'var version := ".*"',
+            to: 'var version := "${nextRelease.version}"',
+            results: [CheckResults("ComAmazonawsKms/src/Index.dfy")],
+            countMatches: true,
+          },
         ],
       },
     ],
@@ -389,6 +398,7 @@ module.exports = {
       {
         assets: [
           "CHANGELOG.md",
+          "ComAmazonawsKms/src/Index.dfy",
           ...Object.values(Runtimes).flatMap((r) => Object.keys(r)),
           ...Object.values(Runtimes.net).flatMap((r) => r.assemblyInfo),
           "**/runtimes/python/**/*.dtr",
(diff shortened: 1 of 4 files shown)


.releaserc.cjs
```diff
@@ -372,6 +372,15 @@ module.exports = {
...
           },
+
+          // Update the version in ComAmazonawsKms/src/Index.dfy DafnyUserAgentSuffix function
+          {
+            files: ["ComAmazonawsKms/src/Index.dfy"],
+            from: 'var version := ".*"',
+            to: 'var version := "${nextRelease.version}"',
+            results: [CheckResults("ComAmazonawsKms/src/Index.dfy")],
+            countMatches: true,
+          },
```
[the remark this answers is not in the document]
rishav-karanjit: tagging a repo is a built in function of semantic release. So, while testing this you might have tagged `v1.11.2`. https://github.com/aws/aws-cryptographic-material-providers-library/tree/v1.11.2. 

You can just delete this tag and its safe because only Go cares about tags and there is no go.mod in the root of the repo. That's the reason we only have tag protection for `releases/go/**/**`.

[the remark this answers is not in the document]
imabhichow: > So, while testing this you might have tagged v1.11.2. https://github.com/aws/aws-cryptographic-material-providers-library/tree/v1.11.2.

Good catch Rishav. I will delete the tag.

ComAmazonawsKms/runtimes/python/src/aws_cryptography_internal_kms/internaldafny/extern/Com_Amazonaws_Kms.py
```diff
@@ -24,15 +26,18 @@ class default__(aws_cryptography_internal_kms.internaldafny.generated.Com_Amazon
     @staticmethod
     def KMSClient(boto_client = None, region = None):
         if boto_client is None:
```
kessplas, line 28: Will this only append the user agent string to the default client (i.e. the client we create when the customer doesn't provide one)? Is it possible to also add it to the user defined one?

imabhichow: Yes, this will only append to the default client. 

>Is it possible to also add it to the user defined one?

I'll look at the Python implementation and see if we can append the user agent. (I assume other runtimes do not append user agent strings to customer-supplied ones).

imabhichow: Oh I think we may be able to append to the existing `user_agent_extra` if there's one.

kessplas (approved):

PR #1844, opened 2026-04-20
chore: update semantic release to differentiate between runtime 

rishav-karanjit:
### Description of changes:

MPL [Java, .NET, Python] will follow same version pattern as MPL Go which means:

- There will be separation in version number of MPL [Java, .NET, Python]. i.e If MPL Java `v1.11.0` then next patch version will be `v1.11.1` even if .NET is in `v2.0.0` and Python is in `v1.11.2`
- [Java, .NET, Python] will have their own CHANGELOG for its own changes only which will be named `CHANGELOG-<runtime>.md`
- The `CHANGELOG.md` will link to `CHANGELOG-<runtime>.md` if its in same directory. For changelog in different directory (like Go it will remain as-is)

Test runs:
1. For Java: [f00a411](https://github.com/aws/aws-cryptographic-material-providers-library/pull/1844/commits/f00a41122c571411c251977214dcfb7d9a33e401)
2. For .NET: [05cc298](https://github.com/aws/aws-cryptographic-material-providers-library/pull/1844/commits/05cc298686063a4003083f0db4c3230f904c379b)
3. For Python: [cc7bc44](https://github.com/aws/aws-cryptographic-material-providers-library/pull/1844/commits/cc7bc44672278a7dadbd1ec8b036ccc0862067b8)

```
chore: update semantic release to differentiate between runtime
```


ComAmazonawsKms/src/Index.dfy
```diff
@@ -30,8 +30,8 @@ module {:extern "software.amazon.cryptography.services.kms.internaldafny"} Com.A
 
   function method DafnyUserAgentSuffix(runtime: string): string
   {
-    // This version is automatically updated by semantic-release
-    var version := "2.0.0";
+    // This version is automatically updated by semantic-release only on Java releases
```
josecorella: why only on java releases? This should get updated on any release going out

project.properties
```diff
@@ -8,4 +8,4 @@
 dafnyVersion=4.9.0
 dafnyVerifyVersion=4.9.1
 dafnyRustVersion=4.10.0
-mplVersion=1.11.2-SNAPSHOT
+javaMPLVersion=1.11.2-SNAPSHOT
```
josecorella, line 11: super nit: you didn't have to change this since java is the only one that uses this variable name

ComAmazonawsKms/src/Index.dfy
rishav-karanjit: Agreed. Although this version is not used today it does not mean we won't use version in user agent string in future. So, putting this is forward looking. 

I updated PR to always update the version.

project.properties
rishav-karanjit: When I was read this, I inferred from variable name that .net and python made use of this version but I only realized after searching the repo that its just Java. So, I changed the variable name to make it more clear

.releaserc-java.cjs
```diff
@@ -0,0 +1,256 @@
...
+  repositoryUrl:
+    "git@github.com:aws/aws-cryptographic-material-providers-library.git",
+  tagFormat: "v${version}-java",
+  plugins: [
+    [
+      "@semantic-release/commit-analyzer",
+      {
+        preset: "conventionalcommits",
+        parserOpts: {
+          noteKeywords: ["JAVA BREAKING CHANGE", "JAVA BREAKING CHANGES"],
```
josecorella, line 40: I would do one more test where you have a NET or Python Breaking Change and check that when you run the java command it does not bump to a major version to ensure that the regex expression is not lazy but strict

josecorella (approved):
lgtm

PR #1858, opened 2026-04-28
feat(ci): Add workflow to auto-sync mutations/mutations with main

sharmabikram:
Add a GitHub Actions workflow that automatically merges main into mutations/mutations on every push to main and on a weekday schedule. If merge conflicts are detected, the workflow notifies the team via Slack using the existing notification infrastructure.

This reduces manual MCM effort and lowers risk during time-sensitive security releases by keeping branch drift minimal.

sim: https://t.corp.amazon.com/P420787080

```
```


.github/workflows/sync-mutations-branch.yml
```diff
@@ -0,0 +1,91 @@
+# This workflow keeps mutations/mutations in sync with main.
+# It creates a PR from main into mutations/mutations, which triggers
+# the existing PR CI. A team member must approve and merge the PR.
+name: Sync mutations/mutations with main
+
+on:
+  push:
+    branches:
+      - main
```
rishav-karanjit:
```suggestion

```

rishav-karanjit: Because we are running it a lot and will create noise.

.github/workflows/sync-mutations-branch.yml
```diff
@@ -0,0 +1,91 @@
...
+  # TODO: Remove pull_request trigger after testing
+  pull_request:
+    branches:
+      - main
+  schedule:
+    # Weekdays at 15:00 UTC (8AM PDT), same cadence as daily CI
```
rishav-karanjit, line 8: Non blocking: What about weekly?

sharmabikram: Less frequent would mean more commits at once meaning more chances of conflict. I would stick with one a day for now and tune if this is too much notification.

.github/workflows/sync-mutations-branch.yml
sharmabikram: This is a temporary code to see how PR will look like which is now reverted.

rishav-karanjit (approved):
LGTM

PR #1872, opened 2026-05-05
chore(CI): split up build from tests to reduce throttling

josecorella:

```
```

### DIFF
--- a/.github/workflows/library_go_build.yml
+++ b/.github/workflows/library_go_build.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Upload generated Go source
         uses: actions/upload-artifact@v7
         with:
-          name: go-generated-${{ matrix.os }}
+          name: go-generated-${{ matrix.os }}-${{ github.sha }}
           path: |
             **/runtimes/go/ImplementationFromDafny-go/
             **/runtimes/go/TestsFromDafny-go/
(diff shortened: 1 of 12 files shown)


rishav-karanjit (commented):
Its a very good idea to cache the generated code. Currently, CI always generates the code in every PR which is redundant and can be resolved with cacheing. 

I have a few question for the diff:

1. I was trying to find if you have skipped build job in `library_<runtime>_build.yml` when generated code is cached but I did not find it. Can you help me by pointing it where it is?
2. From the diff, we have cached the generated code. What happens when one CI which goes off in morning caches generated code and within the retention day someone changes the dafny code later? Will the CI know that the dafny code is changed and updates the cache?

josecorella:
> Its a very good idea to cache the generated code. Currently, CI always generates the code in every PR which is redundant and can be resolved with cacheing.
> 
> I have a few question for the diff:
> 
>     1. I was trying to find if you have skipped build job in `library_<runtime>_build.yml` when generated code is cached but I did not find it. Can you help me by pointing it where it is?
In order to not touch the pull.yml, daily_ci.yml files and the rest the build job gets called from the existing test jobs, build runs first and then test depends on the build one 

>     2. From the diff, we have cached the generated code. What happens when one CI which goes off in morning caches generated code and within the retention day someone changes the dafny code later? Will the CI know that the dafny code is changed and updates the cache?

They are separate, one does not affect the other, but we could add the commit id to the uploaded code if it makes us feel more warm and fuzzy

rishav-karanjit (dismissed):
LGTM. 

The CI is failing. There may some missing installation in the diff. lmk if you need re-approval.

rishav-karanjit (dismissed):

rishav-karanjit (approved):

PR #1875, opened 2026-05-11
chore(ci): add a date timestamp to reduce daily ci flakiness

josecorella:
### Description of changes:
The current retention period for uploaded artifacts is 1 day. (This is the default and you cannot make it hours)
This means that there is a chance that when daily ci runs, the artifacts from the previous days that have the same commit id (no new commits from one day to the next). 
If the action tries to write an entry that already exists daily ci fails with the [error](https://github.com/aws/aws-cryptographic-material-providers-library/pull/1875#issue-4423470378): ``Conflict: an artifact with this name already exists on the workflow run``. To resolve this we add the timestamp to the uploaded artifact which allows for daily ci to run correctly when there has not been prs merged onto master from one day to the next.

```
```

### DIFF
--- a/.github/workflows/library_go_build.yml
+++ b/.github/workflows/library_go_build.yml
@@ -78,7 +78,7 @@ jobs:
         working-directory: ./TestVectorsAwsCryptographicMaterialProviders
         run: |
           make purge_polymorph_code
-      
+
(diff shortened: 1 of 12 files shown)


sharmabikram (dismissed):
Can you please add how this is going to reduce the flakiness in description of this PR?

sharmabikram (approved):

